PR #25, opened 2020-07-27
move continuation return-type to macro (pragma syntax) #20

### LINKED ISSUE #20: move continuation return-type to macro (pragma syntax)

zevv:
Fixes #20. Didn't want to check this in on master because it will probably break your stuff.

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -482,7 +482,7 @@ proc saften(penv: var Env; input: NimNode): NimNode =
       else:
         discard "nil return; no remaining goto for " & $n.kind
 
-macro cps*(n: untyped): untyped =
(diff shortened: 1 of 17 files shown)


cps.nim
```diff
@@ -482,7 +482,7 @@ proc saften(penv: var Env; input: NimNode): NimNode =
       else:
         discard "nil return; no remaining goto for " & $n.kind
 
-macro cps*(n: untyped): untyped =
+macro cps*(T: untyped, n: untyped): untyped =
```
Clyybber, line 485:
```suggestion
macro cps*(T: type, n: untyped): untyped =
```
*should* work too

PR #26, opened 2020-07-28
deep surgery that may well kill the patient

disruptek:


disruptek:
- adds semantics for exceptions
- unifies objects by size _wip, might be done_
- removes name clashes
- adds a result variable
- adds sniffing of the continuation variable name
- installs missing continuation variable as first arg
- closes some issues i guess?

disruptek:
Also, yeah, it's not expected to pass tests or anything. :wink:

zevv:
Can I complain already or should I first properly understand what you did?

Issue #31, opened 2020-08-03
iterative rewrite stages


disruptek:
- [x] add missing returns
- [x] normalize returns
- [x] normalize proc params (3-son identDefs, at least)
- [x] normalize var/let blocks (3-son identDefs, one section per identifier if possible)
- [x] replace stack symbols with continuation symbols
- [x] rewrite hidden conversions
- [x] rewrite unhidden conversions
- [x] replace hidden addr/deref
- [x] normalize tuple deconstruction assignment
- [ ] try exprs
- [x] defer
- [ ] rewrite iterators as cps
- [ ] rewrite cps calls as blocks
- [ ] rewrite cps while as blocks
- [x] rewrite cps break as goto
- [x] rewrite cps continue as goto
- [ ] rewrite cps blocks as procs
- [x] rewrite named break as goto
- [x] rewrite rvalue cps in shims
- [ ] remove gratuitous assignments?  or omit them from earlier stage?
- [x] lift types
- [x] lift procs
- [ ] collapse types

Let's refine this list and add detail, stages...

alaviss:
Pretty sure this can be closed now since we don't rewrite everything in saften anymore.

disruptek:
I've been keeping it accurate and it's no less intentional; what do you want to skip on this list?

PR #49, opened 2021-01-10
conv

zevv:
like so


disruptek:
Added this to #31 also. :+1:

PR #70, opened 2021-04-24
cps: add support for function pointer calls

alaviss:
isCpsCall assumes that the callee is always a nnkSym, which is not the
case when the callee is a function pointer stored inside an object.

Blocked by nim-lang/Nim#17836, but confirmed working with @saem
workaround.

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -26,12 +26,28 @@ when defined(yourdaywillcomelittleonecommayourdaywillcomedotdotdot):
   const
     cpsish = {nnkYieldStmt, nnkContinueStmt}  ## precede cps calls
 
+func getCallSym(n: NimNode): NimNode =
+  ## Get the symbol that is being called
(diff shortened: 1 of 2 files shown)


disruptek:
Nicely done. :+1:

cps.nim
```diff
@@ -26,12 +26,28 @@ when defined(yourdaywillcomelittleonecommayourdaywillcomedotdotdot):
...
+  expectKind n, callish
+  result = n[0]
+  while result != nil:
+    case result.kind
+    of nnkDotExpr:
+      result = result[1]
+    of nnkSym:
+      break
+    of nnkIdent:
+      raise newException(CatchableError, "This call is not typed")
```
disruptek, line 40: Might be useful to betray the identifier that threw this exception, right?  Remember that in macros, the stack trace is meaningless to the programmer.

PR #72, opened 2021-04-26
cps: strip more call nodes types

alaviss:
The previous algorithm has an error:

    Prefix            <- Rewritten
      Sym "not"
      Infix           <- Rewritten
        Sym "=="
        StmtListExpr  <- Not rewritten
          Ident "foo"
        Ident "bar"

Where StmtListExpr is a direct descendant of a call node (Infix) but was
not rewritten.

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -827,21 +827,26 @@ proc cpsXfrmProc(T: NimNode, n: NimNode): NimNode =
       debugEcho repr(result).numberedLines(info.line)
 
 proc workaroundRewrites(n: NimNode): NimNode =
+  proc rewriteContainer(n: NimNode): NimNode =
+    ## Helper function to recreate a container node while keeping all children
+    ## to discard semantic data attached to the container.
+    ##
+    ## Returns the same node if its not a container node.
+    result = n
+    if n.kind notin AtomicNodes:
+      result = newNimNode(n.kind, n)
+      for child in n:
+        result.add child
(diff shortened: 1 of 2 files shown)


cps.nim
```diff
@@ -833,13 +833,14 @@ proc workaroundRewrites(n: NimNode): NimNode =
       # so that sigmatch doesn't decide to skip it
       result = newNimNode(n.kind, n)
       for child in n.items:
+        var newChild = workaroundRewrites child
+        # The containers of direct children always has to be rewritten
```
saem, line 837: Why do the parents of direct children need to be rewritten again?

I keep forgetting why, so I figured it's worth noting the reason here.

saem: Nevermind, github was just hiding the part which explains it. 🤦🏽

saem (approved):

saem:
LGTM

cps.nim
alaviss: Because the compiler sucks :P I figured that I can omit that because this proc is called `workaroundRewrites`.

In more details then it's just that the `.typ` lingers on direct children of call nodes, so we have to rewrite them all the time. I'll add that comment in.

disruptek (approved):

disruptek (approved):

saem (approved):
Ship it like kmart.

PR #75, opened 2021-04-29
cps: remove getCallSym

### LINKED ISSUE #74: More function pointers issue
```nim
import cps

type
  Cont* = ref object of RootObj
    fn*: proc(c: Cont): Cont {.nimcall.}

proc trampoline(c: Cont) =
  var c = c
  while c != nil and c.fn != nil:
    c = c.fn(c)

proc noop*(c: Cont): Cont {.cpsMagic.} = c

proc bar(i: int): proc(): int =
  result = proc(): int =
    i * 2

proc foo() {.cps: Cont.} =
  var i = 2
  noop()
  i = bar(i)()
  noop()
  doAssert i == 4

trampoline foo()
```

Compiler output:

```
=== .cps. on foo(original)  === stretch.nim(18, 0)
 18  proc foo() =
 19    var i = 2
 20    noop()
 21    i = bar(i)()
 22    noop()
 23    echo [i == 4]
 24
nnkVarSection	i: int = 2
stack trace: (most recent call last)
cps.nim(877, 21)         cps
cps.nim(863, 25)         cpsXfrm
cps.nim(799, 15)         cpsXfrmProc
cps.nim(583, 21)         saften
cps.nim(371, 26)         saften
cps.nim(339, 27)         splitAt
cps.nim(320, 15)         procScope
cps.nim(583, 21)         saften
cps.nim(368, 10)         saften
cps.nim(50, 21)          isCpsCall
cps.nim(43, 7)           getCallSym
stretch.nim(18, 14) template/generic instantiation of `cps` from here
cps.nim(43, 7) Error: unhandled exception: unknown node type: nnkCall [Defect]
```

Fixing this is easy, but I might have to rethink the basis of `getCallSym`, since calls in Nim don't just come from symbols, expressions can be called too.

alaviss:
getCallSym() was written on the assumption that all callees are
symbols, which is not the case.

Added the test case that prompted this change.

Fixes #74

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -26,29 +26,14 @@ when defined(yourdaywillcomelittleonecommayourdaywillcomedotdotdot):
   const
     cpsish = {nnkYieldStmt, nnkContinueStmt}  ## precede cps calls
 
-func getCallSym(n: NimNode): NimNode =
-  ## Get the symbol that is being called
-  ## Returns nil if there is no symbol
-  expectKind n, callish
-  result = n[0]
-  while result != nil:
-    case result.kind
-    of nnkDotExpr:
-      result = result[1]
-    of nnkSym:
-      break
-    of nnkIdent:
-      result = nil
-    else:
-      raise newException(Defect, "unknown node type: " & $result.kind)
-
 proc isCpsCall(n: NimNode): bool =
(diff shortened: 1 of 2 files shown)


disruptek (approved):
:shipit:

PR #77, opened 2021-04-29
cps: keep going if there is somewhere to go

alaviss:
Unless you're in a loop.

This is a WIP.

Don't trust this stuff, shadow impossible test still not working since the last three splits in the block got skipped


alaviss:
#85 replaced this

Issue #80, opened 2021-05-02
Incorrect rewrite for defer


alaviss:
```nim
import cps

type
  InfiniteLoop = CatchableError
  Cont* = ref object of RootObj
    when cpsMutant:
      fn*: proc(c: var Cont) {.nimcall.}
    else:
      fn*: proc(c: Cont): Cont {.nimcall.}

var jumps: int

proc trampoline(c: Cont) =
  jumps = 0
  var c = c
  while c != nil and c.fn != nil:
    c = c.fn(c)
    inc jumps
    if jumps > 1000:
      raise newException(InfiniteLoop, $jumps & " iterations")

proc noop*(c: Cont): Cont {.cpsMagic.} = c

template check(cond: bool, msg: string) =
  if not cond:
    raise newException(Defect, "not " & astToStr(cond) & ": " & msg)

var r = 0

proc foo() {.cps: Cont.} =
  defer:
    check r == 2, "defer run before end of scope"
    inc r

  inc r
  noop()
  inc r

trampoline foo()
check r == 3, "expected 3, got: " & $r
```

Current output:

```
Error: unhandled exception: not r == 2: defer run before end of scope [Defect]
```

disruptek:
As far as where `defer` actually ends up, it's a bit of a head-scratcher.  The best solution is probably to rewrite it in a very early pass as we do `for`.  Did you happen to try using the same `getTransformImpl` hack?

disruptek:
This is stalled pending a compiler bug which @alaviss may decide to link here. :kissing_heart:

alaviss:
blocked by https://github.com/nim-lang/Nim/issues/18059

Issue #78, opened 2021-04-29
Splitting inside an if-else generates broken ast


alaviss:
```nim
import cps

type
  Cont* = ref object of RootObj
    fn*: proc(c: Cont): Cont {.nimcall.}

proc trampoline(c: Cont) =
  var c = c
  while c != nil and c.fn != nil:
    c = c.fn(c)

proc noop*(c: Cont): Cont {.cpsMagic.} = c

template check(b: bool, msg: string) =
  if not b:
    raise newException(AssertionDefect, msg)

var r = 0
proc b() {.cps: Cont.} =
  if true:
    noop()
    inc r
  else:
    discard
  inc r

trampoline b()
check r == 2, "expected 2, got: " & $r
```

Output:

```
=== .cps. on b(original)  === temp.nim(19, 0)
 19  proc b() =
 20    if true:
 21      noop()
 22      inc r, 1
 23    else:
 24      discard
 25    inc r, 1
 26
storing type env_402653455
next type env_402653738
=== .cps. on b(transform) === temp.nim(19, 0)
 19
 20  type
 21    env_402653455 = ref object of Cont
 22
 23  proc done_402653695(continuation: Cont): Cont {.cpsLift.}
 24  proc done_402653705(continuation: Cont): Cont {.cpsLift.}
 25  proc done_402653706(continuation: Cont): Cont {.cpsLift.}
 26  proc done_402653707(continuation: Cont): Cont {.cpsLift.}
 27  proc afterCall_402653708(continuation: Cont): Cont {.cpsLift.}
 28  proc done_402653695(continuation: Cont): Cont {.cpsLift.} =
 29    ## saften at 20 of temp.nim
 30    inc r, 1
 31
 32  proc done_402653705(continuation: Cont): Cont {.cpsLift.} =
 33    ## saften at 688 of cps.nim
 34    inc r, 1
 35
 36  proc done_402653706(continuation: Cont): Cont {.cpsLift.} =
 37    ## saften at 20 of temp.nim
 38    else:
 39      discard
 40
 41  proc done_402653707(continuation: Cont): Cont {.cpsLift.} =
 42    ## saften at 688 of cps.nim
 43    else:
 44      discard
 45
 46  proc afterCall_402653708(continuation: Cont): Cont {.cpsLift.} =
 47    ## saften at 871 of cps.nim
 48    inc r, 1
 49
 50  proc b(continuation: Cont): Cont {.cpsCall.} =
 51    ## saften at 19 of temp.nim
[...]

Issue #76, opened 2021-04-29
Block rewriting error when there are no explicit break


alaviss:
Found while messing with "shadow mission impossible"

```nim
import cps

type
  Cont* = ref object of RootObj
    fn*: proc(c: Cont): Cont {.nimcall.}

proc trampoline(c: Cont) =
  var c = c
  while c != nil and c.fn != nil:
    c = c.fn(c)

proc noop*(c: Cont): Cont {.cpsMagic.} = c

template doAssert(b: bool, msg: string) =
  if not b:
    raise newException(AssertionDefect, msg)

var r = 0
proc b() {.cps: Cont.} =
  inc r
  block:
    noop()
    echo "in block"
    inc r
  inc r
  echo "out block"

trampoline b()
doAssert r == 3, "expected 3, got: " & $r
```

Output:

```
=== .cps. on b(original)  === temp.nim(19, 0)
 19  proc b() =
 20    inc r, 1
 21    block:
 22      noop()
 23      echo ["in block"]
 24      inc r, 1
 25    inc r, 1
 26    echo ["out block"]
 27
storing type env_385876243
next type env_385876522
=== .cps. on b(transform) === temp.nim(19, 0)
 19
 20  type
 21    env_385876243 = ref object of Cont
 22
 23  proc blockBreak_385876482(continuation: Cont): Cont {.cpsLift.}
 24  proc afterCall_385876492(continuation: Cont): Cont {.cpsLift.}
 25  proc blockBreak_385876482(continuation: Cont): Cont {.cpsLift.} =
 26    ## saften at 21 of temp.nim
 27    inc r, 1
 28    echo "out block"
 29
 30  proc afterCall_385876492(continuation: Cont): Cont {.cpsLift.} =
 31    ## saften at 848 of cps.nim
 32    echo "in block"
 33    inc r, 1
 34
 35  proc b(continuation: Cont): Cont {.cpsCall.} =
 36    ## saften at 19 of temp.nim
 37    ## saften at 20 of temp.nim
 38    inc r, 1
 39    block:
 40      ## saften at 22 of temp.nim
 41      ## re-use the local continuation by setting the fn
 42      continuation.fn = afterCall_385876492
 43      return noop(continuation)
 44    ## add tail call for block-break proc
 45    ## verbatim tail call
 46    ## new tail call: blockBreak_385876482
 47    ## re-use the local continuation by setting the fn
 48    continuation.fn = blockBreak_385876482
 49    return continuation
[...]

PR #81, opened 2021-05-02
a bunch of failing tests as refactoring prep

alaviss:
Ref #76 
Ref #78 
Ref #80
Ref #56
Ref #51

### DIFF
--- a/tests/taste.nim
+++ b/tests/taste.nim
@@ -816,3 +816,38 @@ suite "tasteful tests":
 
     trampoline foo()
     check r == 2
+
+  block:
+    ## implicit generics
+    when true:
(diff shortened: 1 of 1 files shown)


tests/taste.nim
```diff
@@ -656,3 +656,198 @@ suite "tasteful tests":
...
+      defer:
+        check r == 2, "defer run before end of scope"
+        inc r
+
+      inc r
+      noop()
+      inc r
+
+    trampoline foo()
+    check r == 2
```
disruptek, line 818: Off by one, right?
```suggestion
    check r == 3
```

disruptek (changes requested):
:+1:

PR #84, opened 2021-05-03
cps: experimental support for for-loops

alaviss:
For-loops are inlined using block statements, which messes with their
control flow due to #76.

Breaks "for loop with continue, break" test. On the flip side,
splitting now functions on them.

Also breaks a test in tzevv due to defer being rewritten into
try-finally, which doesn't work due to #80.


cps.nim
```diff
@@ -603,7 +603,12 @@ proc normalizingRewrites(n: NimNode): NimNode =
       if n.len == 2:
         n.add newEmptyNode()
       elif n[1].isEmpty:          # add explicit type symbol
-        n[1] = getTypeInst n[2]
+        if not n[2].isEmpty:
+          n[1] = getTypeInst n[2]
+        else:
+          # get the type from the symbol as the last resort.
+          # walkaround for #48.
```
disruptek, line 610: still workaround

cps.nim
```diff
@@ -854,13 +870,21 @@ proc cpsXfrm(T: NimNode, n: NimNode): NimNode =
     result = copy n
   result = workaroundRewrites(result)
 
-macro cps*(T: typed, n: typed): untyped =
+macro cps2(T: typed, n: typed): untyped =
   # I hate doing stuff inside macros, call the proc to do the work
```
disruptek, line 874:
```suggestion
  ## We used `getImplTransformed` to make a pre-pass whatfer transforming `for` and `defer`
  ## into the input to this macro, where we perform the meat of the CPS transform.
```

cps.nim
```diff
@@ -854,13 +870,21 @@ proc cpsXfrm(T: NimNode, n: NimNode): NimNode =
...
   when defined(nimdoc):
     result = n
   else:
     result = cpsXfrm(T, n)
```
disruptek, line 878: This `when` should move into `cps()`.

cps.nim
```diff
@@ -854,13 +870,21 @@ proc cpsXfrm(T: NimNode, n: NimNode): NimNode =
...
 
+macro cps*(T: typed, n: typed): untyped =
```
disruptek, line 880:
```suggestion
macro cps*(T: typed, n: typed): untyped =
  ## Rewrite procedure `n` in Continuation-Passing Style, extending type `T` to store any transient locals.
```

disruptek (changes requested):
Nice! :+1:

cps.nim
saem: I think we could use a for loop macro to remove part of the getImplTransformed dependency, I still need to think about the defer one.

saem (approved):
Noted the for loop macro idea for posterity. @disruptek already highlighted the doc bits, lgtm.

alaviss:
Rebased, compiler issues that I found:

- `case` statement without `else` will have an `else: nil` appended:

```nim
case true
of true:
  echo "true"
of false:
  echo "failed"
```

transformed to:

```nim
case true
of true:
  echo "true"
of false:
  echo "failed"
else:
  nil
```

And this stuff don't pass sem, obviously.

- the flattening of string concatenation:

```nim
a & b & c
```

transformed to

```
`&`(a, b, c)
```

And it also doesn't pass sem because a `&` with 3 arguments doesn't exist.

I'd say we should perform this transform ourselves.

PR #85, opened 2021-05-06
[WIP] move things into different passes

alaviss:
This thing is made of hacks, don't use, just look.

`scratch.nim` is the simple test that compile.

`taste` won't compile.


cps.nim
```diff
@@ -327,6 +327,138 @@ proc splitAt(env: var Env; n: NimNode; i: int; name = "splat"): Scope =
...
+
+proc makeContProc(name, cont, body: NimNode): NimNode =
+  ## creates a continuation proc from with `name` using continuation `cont`
+  ## with the given body.
+  let
+    body =
+      if body.kind != nnkStmtList:
+        newStmtList body
+      else:
+        body
```
disruptek, line 340:
```suggestion
  # we always wrap the body because there's no reason not to
  let body = newStmtList body
```

cps.nim
```diff
@@ -327,6 +327,138 @@ proc splitAt(env: var Env; n: NimNode; i: int; name = "splat"): Scope =
...
+  expectKind cont, nnkSym
+
+  debug("cpsMayJump", n, akOriginal)
+  debug("cpsMayJump", after, akOriginal)
+
+  var n =
+    if n.kind != nnkStmtList:
+      newStmtList n
+    else:
+      n
```
disruptek, line 440:
```suggestion
  # we always wrap the input because there's no reason not to
  var n = newStmtList n
```

cps.nim
```diff
@@ -351,40 +483,67 @@ proc saften(parent: var Env; n: NimNode): NimNode =
...
+      when false:
+        # if the child is a cps block (not a call), then push a tailcall
+        # onto the stack during the saftening of the child
+        if nc.kind notin unexiter and nc.isCpsBlock and not nc.isCpsCall:
+          withGoto env.splitAt(n, i, "done"):
+            result.doc "saftening during done"
+            result.add env.saften(nc)
+            # we've completed the split, so we're done here
+            return
+      else:
```
disruptek, line 523: Let's just get rid of it.

cps.nim
```diff
@@ -351,40 +483,67 @@ proc saften(parent: var Env; n: NimNode): NimNode =
...
+          tailCall(env, env.saften nc):
+            returnTo after
+        # include the definition for the after proc
+        if after.node.kind != nnkSym:
+          # add the proc definition and declaration without the return
+          for child in after.node.items:
+            if child.kind == nnkProcDef:
+              result.add child
+        # done!
+        return
```
disruptek, line 511: Let's just get rid of it.

cps.nim
```diff
@@ -659,21 +767,71 @@ proc cloneProc(n: NimNode): NimNode =
...
+    of nnkPragma:
+      # a {.cpsPending.} annotation is a standalone pragma statement
+      if n.len == 1 and n.hasPragma("cpsPending"):
+        if replacement.isNil:
+          result = newEmptyNode()
+        else:
+          result = copyNimTree replacement
+    else: discard
+
+  result = filter(n, resolved)
```
disruptek, line 785: Prefer an additional version that takes no replacement argument and does the right thing, versus a version that has special logic to handle a sentinel `nil`.  The current version should demand that the replacement is not nil.

cps.nim
```diff
@@ -659,21 +767,71 @@ proc cloneProc(n: NimNode): NimNode =
...
+  ##
+  ## this is not needed, but it's here so we can change this to
+  ## a sanity check pass later.
+  expectKind n, nnkProcDef
+  debug(".cpsStripPending.", n, akOriginal)
+
+  # make `n` safe for modification
+  let n = normalizingRewrites n
+  # replace all `pending` with `return nil`, signifying end of continuations
+  result = replacePending(n, tailCall(nil, nil))
```
disruptek, line 798: Prefer not providing `replacePending` a second argument, but failing that...
Prefer a `tailCall` that does the right thing with no input...  same as my prior feedback; this is very smelly.  :grin:

cps.nim
```diff
@@ -659,21 +767,71 @@ proc cloneProc(n: NimNode): NimNode =
...
+  result = workaroundRewrites result
+
+  debug(".cpsStripPending.", result, akOriginal, n)
+
+proc xfrmFloat(n: NimNode): NimNode =
```
disruptek, line 803: Maybe another `rewrite` name like `rewriteLift` is more obvious?

tests/scratch.nim
```diff
@@ -0,0 +1,43 @@
...
+  inc r
+  if true:
+    noop()
+    inc r
+    noop()
+    inc r
+  inc r
+
+trampoline foo()
+check r == 4
```
disruptek, line 43: If you really want it, I'd prefer just moving this test into `tzevv` or `taste` to reduce the matrix.

cps/spec.nim
```diff
@@ -143,15 +118,55 @@ when cpsDebug:
...
+        if info.isNil:
+          n
+        else:
+          info
+      lineInfo = info.lineInfoObj
+      procName =
+        if info.kind in RoutineNodes:
+          repr info.name
+        else:
+          ""
```
disruptek, line 139: Feel free to make this 3-4 lines.

disruptek:
Looks great and I hope we don't have to hold this up with getting anything more than `tzevv` blocks to pass.  There's already a newly-working `taste` test and this is already good work we shouldn't have to review twice. :grin:

tests/scratch.nim
alaviss: It's just a sample file so I can keep track of what's compiling.

Now that `taste` is compiling, I will retire this.

cps.nim
```diff
@@ -451,78 +452,39 @@ macro cpsMayJump(cont, n, after: typed): untyped =
...
-          if n[0][1] == label or n[0][1].kind == nnkNilLit:
-            # {.cpsBreak: label.} or {.cpsBreak(nil).}
-            tail
-          else:
-            # {.cpsBreak: not-our-label.}
-            nil
-        else:
-          # {.cpsBreak.}
-          tail
+func matchCpsBreak(label: NimNode): Matcher =
```
saem, line 454: matcherCpsBreak, since it creates one rather than matches it? Not a blocker, cpsBreakMatcher reads better but the lack of symmetry sucks for grouping related things. I can do this later too.

disruptek:
Really excellent work, @alaviss! :kissing_heart:

PR #86, opened 2021-05-10
rm cpsMutant, cpsCast

disruptek:


alaviss:
If it compiles, LGTM. What was the idea behind `cpsMutant` though?

alaviss (approved):
LGTM, we can bring it back if we ever need it.

saem:
Strongly agree, can be resurrected pretty easily down the road.

PR #87, opened 2021-05-10
cps: rewrite defer to try-finally

alaviss:
The properties described in
https://nim-lang.org/docs/manual.html#exception-handling-defer-statement
appears to be preserved correctly.

~~This breaks `tzevv`'s "defer" test, since we don't handle try-finally.~~ No longer happening with the new approach.

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -886,7 +886,7 @@ proc cpsXfrmProc(T: NimNode, n: NimNode): NimNode =
   n.body = env.prepProcBody(newStmtList n.body)
 
   # transform defers
-  n.body = xfrmDefer n.body
+  n.body = rewriteDefer n.body
 
   # ensaftening the proc's body
   n.body = env.saften(n.body)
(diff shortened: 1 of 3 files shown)


alaviss:
> This impl looks too complex to me; for example, I don't see why mutation should occur when no `defer` exists in the input.

If there are no defer then this transform just flatten the stmtlists, so there are no harm afaict.

We can have a "hasDefer" proc that perform the check beforehand if you want it, ofc.

saem (dismissed):
I don't see how to simplify this PR, and it's not a big enough issue to block, IMO.

disruptek:
> If there are no defer then this transform just flatten the stmtlists, so there are no harm afaict.
> 
> We can have a "hasDefer" proc that perform the check beforehand if you want it, ofc.

Makes sense to me; I am not a fan of relying on this code's assumptions for every single transform.

PR #88, opened 2021-05-11
cps: make cpsMagic proc raises Defect when used outside {.cps.}

alaviss:
Currently we put a `{.warning.}` inside the proc which does... absolutely
nothing but generates an useless warning.

Options to catch this at compile-time are non-existant, unfortunately,
since the procs using these cpsMagic procedures must sem before
they can be processed by `{.cps.}`.

So instead we are gonna make these cpsMagic proc raises when the
"non-cps" version is used outside of `{.cps.}`. Should help us catches
bugs in the transform if we misses anything.

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -959,11 +959,19 @@ macro cpsMagic*(n: untyped): untyped =
   when false:
     m.addPragma newColonExpr(ident"error", msg.newLit)
     m.body.add nnkDiscardStmt.newNimNode(n).add newEmptyNode()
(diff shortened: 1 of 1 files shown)


disruptek:
How do you propose to write this code?

```nim
spawn foo()
```

`spawn()` is a `.cpsMagic` that simply adds a continuation to the `eventqueue` dispatcher; it's effectively the same as `jield` IIRC, but with a familiar name.

PR #89, opened 2021-05-12
cps/spec: rework varargs rewrite

alaviss:
Mostly moving code around, but the difference is that brackets
are now kept when the node being rewritten is not the last node
or when the call node is a binary/unary call.

Fixes the issue @disruptek sent on IRC and a bunch more.


disruptek:
Nice; the tests are 10x the value of the code.  :kissing_heart:

PR #94, opened 2021-05-13
cps: add limited exception handling support

alaviss:
Experimental and not fully tested yet

Splitting inside any `except` or `finally` bodies are
not supported and will yield an error

/cc @disruptek I need you to break this


cps.nim
```diff
@@ -514,6 +516,101 @@ macro cpsWhile(cont, cond, n: typed): untyped =
...
+  proc annotateOk(n: NimNode): NimNode =
+    ## Add `ok = true` before any control flow that stays in the current scope
+    ##
+    ## Control flow exiting the current scope should be unresolved at this
+    ## stage and we shouldn't have to care about them
+    case n.kind
+    of nnkProcDef:
+      if not n.hasPragma("cpsContinuation"):
+        result = n
+    of nnkReturnStmt, nnkBreakStmt, nnkContinueStmt:
```
alaviss, line 599: We only have to handle `nnkReturnStmt`, any `nnkBreakStmt` or `nnkContinueStmt` should be within the same continuation, thus `finally` won't trigger.

cps.nim
```diff
@@ -514,6 +516,101 @@ macro cpsWhile(cont, cond, n: typed): untyped =
...
+    okSym = desym genSym(nskParam, "ok")
+    handlerWrapperBody = newStmtList()
+
+  # mixin `cont`
+  handlerWrapperBody.add:
+    nnkMixinStmt.newTree:
+      desym cont
+
+  # var ok {.inject.} = false
+  handlerWrapperBody.add newVarStmt(
```
alaviss, line 547: We only need this stuff if there is a `finally` body, since it is an indicator for whether the finally branch should be executed

cps.nim
disruptek: As discussed, `finally` is rewritten into `block`, right?

alaviss: Nah, I'm just replicate it across all legs, with a tag letting us know whether the finally should be run. There might be a bug or two with that, but I haven't figured out any tests yet.

disruptek: The `finally` clause should always be run; that's the point.  When you rewrite it into `block` statements, you produce two (or more) control-flow paths and all receive a copy of the `finally` code -- perhaps via a template.  Any subsequent rewrite deals with these different paths via our normal `block`-handling logic, so the `finally` effectively evaporates.

It's kinda like your solution to `if` expressions...  You dig down into the logic and say, "okay, now what are we really trying to say here..." and that statement is "treat this as an expression on the RHS of an assignment".

In this case, what we are really trying to say is, "do this stuff regardless of how you exit the scope".

PR #98, opened 2021-05-15
Refactoring and cleanups

zevv:
because less is more

This is mostly moving and simplifying code, cleanups, some comment updates and throwing out old blocks of code wrapped in `when false`

Also removing code in old conditionals which is all broken at this time

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -56,10 +56,7 @@ proc addReturn(p: var NimNode; n: NimNode) =
       elif not n.firstReturn.isNil:
         n
       else:
-        when cpsMoves:
-          nnkReturnStmt.newNimNode(n).add newEmptyNode()
-        else:
-          nnkReturnStmt.newNimNode(n).add n
+        nnkReturnStmt.newNimNode(n).add n
   else:
     p.doc "omitted a return of " & repr(n)
 
@@ -148,7 +145,7 @@ proc cmpKind(a, b: NimNode): int =
     result = -1
 
 proc lambdaLift(lifted: NimNode; n: NimNode): NimNode =
-  ## lift ast tagged with cpsLift pragma to top-level and omit the pragma
+  ## lift AST tagged with cpsLift pragma to top-level and omit the pragma
 
   # pull the liftable procs out of the input; the result is, uh, `result`
   proc liften(n: NimNode): NimNode =
@@ -705,7 +702,7 @@ proc saften(parent: var Env; n: NimNode): NimNode =
     else:
       result.add env.saften(nc)
 
-proc cloneProc(n: NimNode): NimNode =
+proc cloneProc(n: NimNode, body: NimNode = nil): NimNode =
   ## create a copy of a typed proc which satisfies the compiler
   assert n.kind == nnkProcDef
   result = nnkProcDef.newTree(
@@ -715,7 +712,7 @@ proc cloneProc(n: NimNode): NimNode =
     n.params,
     newEmptyNode(),
     newEmptyNode(),
-    copy n.body)
+    if body == nil: copy n.body else: body)
 
 proc replacePending(n, replacement: NimNode): NimNode =
   ## Replace cpsPending annotations with something else, usually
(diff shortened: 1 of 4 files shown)


cps.nim
```diff
@@ -787,74 +787,48 @@ macro cpsFloater(n: typed): untyped =
...
-
-    booty = cloneProc n
-    booty.body = newStmtList()
-
-    # add the remaining proc params to the environment
-    # and set them up in the bootstrap at the same time
-    for defs in n.params[first .. ^1]:
-      for name, list in env.localSection(defs):
-        preamble.add list
+  ## XXX I do not understand what this does. If I take this out, I get `Error:
```
zevv, line 808: Can someone explain to me why this is? `env.localSections()` probably has side effects I do not understand yet.

cps.nim
```diff
@@ -787,74 +787,48 @@ macro cpsFloater(n: typed): untyped =
...
+  var booty = cloneProc(n, newStmtList())
+  booty.params[0] = T
+  booty.body.add doc "This is the bootstrap to go from Nim-land to CPS-land"
+  booty.body.add newAssignment(ident"result", env.newContinuation(env.first, booty.name))
 
   # we can't mutate typed nodes, so copy ourselves
   n = cloneProc n
 
-  # and do some pruning of these typed trees
+  # do some pruning of these typed trees. XXX Why? 7?
```
zevv, line 823: Yeah, why?

cps.nim
```diff
@@ -895,70 +858,38 @@ proc cpsXfrmProc(T: NimNode, n: NimNode): NimNode =
   # spamming the developers
   debug(".cps.", result, Transformed)
 
-proc cpsXfrm(T: NimNode, n: NimNode): NimNode =
```
zevv, line 898: We used to have cps top level blocks one day, but this is no longer a thing and was broken anyway. We can always re-add this later.

cps.nim
```diff
@@ -895,70 +858,38 @@ proc cpsXfrmProc(T: NimNode, n: NimNode): NimNode =
...
 
-  # ensure that .cpsCall. is added to the copies of the proc
+  # Add .cpsCall. pragma to the proc
   n.addPragma ident"cpsCall"
 
-  # create a version of the proc that pukes outside of cps context
+  # create a Nim-land version of the proc that throws an exception when called
+  # from outside of CPS-land. XXX We also need this to keep the nim compiler
+  # happy, otherwise we get a type mismatch on call. Do we need a special name
+  # for this type of proc?
```
zevv, line 879: Input on the above appreciated

cps.nim
```diff
@@ -705,7 +705,7 @@ proc saften(parent: var Env; n: NimNode): NimNode =
     else:
       result.add env.saften(nc)
 
-proc cloneProc(n: NimNode): NimNode =
+proc cloneProc(n: NimNode, body: NimNode = nil): NimNode =
```
zevv, line 708: There is a few places where we first copy the body, and then replace it with something else. This saves the copy.

cps.nim
alaviss: pretty sure it's to de-sem the thing. sem-ed procs has a special result field (at index 7) so we need to get rid of it, not sure if desym is necessary though

cps.nim
alaviss: The iterator itself adds params to the environment. You see environment misses errors because the params weren't moved into the continuation

cps.nim
disruptek: Right; the name is a symbol and we know it's going to change, so we desym it to make sure we don't refer to the wrong symbol later accidentally.

cps.nim
disruptek: We could expose the side-effects, but it just moves complexity into multiple other spots.

cps.nim
disruptek: Honestly, I don't think it should throw an exception outside; the library author may have a legitimate reason to use a procedure like `sleep` or `enqueue` on a continuation.  What about creating a `warning` outside of `release` builds?

cps/environment.nim
```diff
@@ -148,12 +148,6 @@ proc isDirty*(e: Env): bool =
   else:
     result = not(e.isWritten) or (not(e.parent.isNil) and e.parent.isDirty)
 
-proc setDirty(e: var Env) =
-  when false:
-    assert not e.isNil
-    e.id = newEmptyNode()
-    assert e.isDirty
-
```
disruptek, line 156: We will need this stuff if we ever want to support multiple object types according to position in the continuation; I know it seems like a long shot at the moment, but it might be worth resurrecting behind a flag to help prevent bitrot.

disruptek (commented):
It's fair, but I'd say if you're going to go down this path, you could remove or replace `scopes.nim`.

@alaviss added the data in the form of AST annotations, now we need to query those instead of tracking `Scope` manually.  This would also let us compose the environment in a single, separated pass at the end, which safely discards the `dirty` flag since any future impl will work differently.

cps.nim
alaviss: The version outside of CPS do nothing, and doesn't even take a continuation.

zevv: Right, it only exists so Nim is able to compile the code in the first place

alaviss (approved):
Less code less bugs

PR #101, opened 2021-05-16
several issues with tuple deconstruction

disruptek:
This solves a couple problems with tuples that mostly manifest like this:

```nim
let (a, b, c) = foo()
```
is now
```nim
(env.a, env.b, env.c) = foo()
```
...and there was something wacky with distinct types in tuple deconstruction that was more trivial.

Also, it simplifies `localSection` to no longer iterate, `addAssignment` to yield only one value, and removes some other dead code in `initialization`.

Also comments out all but the "top-level" debugging output.

Also adds a procedure for fetching a single LHS symbol for assignment in the shim codepath, which I don't think we actually use.  It made more sense to do this than complicate the other aforementioned procs...

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -366,7 +366,7 @@ macro cpsJump(cont, call, n: typed): untyped =
   expectKind cont, nnkSym
   expectKind call, nnkCallKinds
 
-  debug("cpsJump", n, Original)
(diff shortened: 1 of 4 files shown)


saem (approved):
I tried hurting it with bad ideas and it kept working.

Imma call it good.

PR #103, opened 2021-05-16
complain on var parameters

zevv:
untile one day we might support boxing.

or maybe error instead?

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -804,6 +804,8 @@ proc cpsXfrmProc(T: NimNode, n: NimNode): NimNode =
 
   # add parameters into the environment
   for defs in n.params[1 .. ^1]:
+    if defs[1].kind == nnkVarTy: 
+      error "cps does not support var parameters", n 
     for _, _ in env.localSection(defs):
       discard
 


cps.nim
```diff
@@ -804,6 +804,8 @@ proc cpsXfrmProc(T: NimNode, n: NimNode): NimNode =
 
   # add parameters into the environment
   for defs in n.params[1 .. ^1]:
+    if defs[1].kind == nnkVarTy: 
+      warning "cps does not support var parameters", n
```
disruptek, line 808: I'm fine with it; alternatively, maybe
```suggestion
    if defs[1].kind == nnkVarTy:
      warning "var paramerters behavior is undefined in cps", n
```
...since there are technically scenarios in which it will work.

zevv: Yeah, that's called undefined behavior. Let's not go there, shall we. Either we support it or we don't, but if we go doing "maybe" we end up with crappy code and people will complain if we change the UB  later.
Instead of a warning, I'd even prefer making it an error.

zevv: I made it an error.

disruptek: Fair.

PR #105, opened 2021-05-16
Added explict state enum and query proc + tests

zevv:


cps.nim
```diff
@@ -8,6 +8,14 @@ import cps/environment
 export Continuation, ContinuationProc, cpsCall
 export cpsDebug, cpsTrace
 
+type
+
+  cpsState* = enum
+    ## Representation of the state of a continuation.
+    Running,   ## The continuation is active and running and can be resumed
```
disruptek, line 15: One might measure the state of a continuation from inside the continuation, or from a dispatcher that is contemplating resuming the continuation.  I guess the point is, it's easier to define `Running` as `not (Finished or Dismissed)`.

Also, maybe the sort order should change to `Dismissed`, `Running`, `Finished` (or the reverse)?

cps.nim
```diff
@@ -854,9 +862,18 @@ proc cpsXfrmProc(T: NimNode, n: NimNode): NimNode =
...
   debug(".cps.", result, Transformed)
+    
+proc state*(c: Continuation): cpsState =
+  ## Get the current state of a continuation
+  if c == nil:
+    result = Dismissed
+  elif c.fn == nil:
+    result = Finished
+  else:
+    result = Running
```
disruptek, line 873: just omit `result =` so it's impossible to not set it correctly, right?

PR #108, opened 2021-05-18
Moved cps xfrm code to own file, toplevel cps.nim now only contains p…

zevv:
…ublic API


disruptek:
What's the point?  And why the insistance on `xfrm`?  Is `transform` really that hard to type?

zevv:
Personal preference I guess: top level cps.nim contains all that is public, and when we start properly documenting this is where all the text should go.

Everything under the `/cps` dir is implementation - internals that are of no business for nosy lib readers.

Feel free to close and throw away, of course.

PR #116, opened 2021-05-21
formalized hooks: trace, alloc, dealloc, coop

disruptek:
By formalized, I think @zevv just wants to make the stuff discoverable and not too adhoc.  This does that, but I'm not sure how useful it is.

I think the `trace()` might want to take the continuation as input as well, so that could mess with the invocation somewhat.  Maybe we can standardize on just a few signatures for hooks...

### DIFF
--- a/tests/taste.nim
+++ b/tests/taste.nim
@@ -1135,12 +1135,15 @@ suite "tasteful tests":
     ## custom deallocators
     var r = 0
     proc dealloc(c: Cont) =
+      check r == 0
       inc r
 
     proc foo(x: int) {.cps: Cont.} =
+      check r == 0
       check x == 3
       noop()
       check x == 3
+      check r == 0
 
     trampoline foo(3)
     check r == 1, "bzzzt"
--- a/tests/taste.nim
+++ b/tests/taste.nim
@@ -1,7 +1,7 @@
+import std/macros
+import std/strutils
 import balls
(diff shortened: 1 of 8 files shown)


disruptek:
Of course, when I tried to actually use allocators, it blew up with the following problems:
- `typedesc` is such an opaque value that it's hard to actually work with it; you cannot compare it to other typedescs, you cannot hash it, you can get the `sizeof` but this is useless for `ref` types, etc.
- `$typedesc` on our continuation types yields `env`, which is not terribly useful

tests/taste.nim
```diff
@@ -1090,3 +1090,60 @@ suite "tasteful tests":
...
+                     "whileLoop", "24", "16", "afterCall", "8", "16",
+                     "whileLoop", "24", "16", "afterCall", "8", "16",
+                     "whileLoop", "24", "16", "afterCall", "8", "16", ]
+
+  block:
+    ## custom allocators
+    var r = 0
+    proc alloc[Cont](c: typedesc[Cont]): c =
+      inc r
+      new c
```
saem, line 1124: Try using Cont, the whole typedesc thing is butts.

See: https://github.com/nim-lang/Nim/blob/e12597589f2a8cdaabe09e8b2b92ff0fbe3727ef/lib/system.nim#L978

PR #121, opened 2021-05-22
Fix trampoline

zevv:


zevv:
or should we explicitly `sink` it?

PR #123, opened 2021-05-24
first half of parentage including whelp

disruptek:
What's cool about this is that it makes CPS _colorless by default_ and the user only adds color when they wish to exploit the concurrency features of continuations without the support of the library author.

This might be a little sloppy because things came together so quickly.

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -77,7 +77,6 @@ macro cpsMagic*(n: untyped): untyped =
   when not defined(nimdoc):
     result.add n
   result.add m
-  echo result.repr
 
 proc doWhelp(n: NimNode; args: seq[NimNode]): NimNode =
   for n in n.pragma.items:
--- a/cps.nim
+++ b/cps.nim
@@ -76,11 +76,11 @@ macro cpsMagic*(n: untyped): untyped =
     result.add n
   result.add m
 
-proc doWhelp(n: NimNode): NimNode =
+proc doWhelp(n: NimNode; args: seq[NimNode]): NimNode =
   for n in n.pragma.items:
     if n.kind == nnkExprColonExpr:
       if $n[0] == "cpsBootstrap":
-        result = newCall(n[1])
+        result = n[1].newCall args
   if result.isNil:
     result = n.errorAst "welping malfunction"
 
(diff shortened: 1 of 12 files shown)


cps.nim
```diff
@@ -76,6 +76,23 @@ macro cpsMagic*(n: untyped): untyped =
     result.add n
   result.add m
 
+proc doWhelp(n: NimNode; args: seq[NimNode]): NimNode =
+  for n in n.pragma.items:
+    if n.kind == nnkExprColonExpr:
+      if $n[0] == "cpsBootstrap":
+        result = n[1].newCall args
```
alaviss, line 83: Need a new `getPragma` proc, will be useful to inspect other annotations.

Maybe we should get @Clyybber to expose more of the stuff that's already in `std/macros`.

cps/environment.nim
```diff
@@ -441,3 +444,53 @@ proc getException*(e: var Env): NimNode =
...
+proc createWhelp*(env: Env; n, goto: NimNode): NimNode =
+  ## the whelp needs to create a continuation
+  result = cloneProc(n, newStmtList())
+  result.params[0] = env.root
+  result.name = nskProc.genSym"whelp"
+  result.introduce {Alloc}
+
+  # create the continuation as the result and point it at the proc
+  result.body.add:
+    env.createContinuation(ident"result", desym goto)
```
alaviss, line 457: Why are we de-syming `goto`?

cps/environment.nim
```diff
@@ -441,3 +444,53 @@ proc getException*(e: var Env): NimNode =
...
+  # create the continuation using the new variable and point it at the proc
+  result.body.add:
+    env.createContinuation(c, desym goto)
+
+  # now the trampoline
+  result.body.add:
+    nnkWhileStmt.newTree: [
+      newCall(ident"running", c),  # XXX: bindSym?  bleh.
+      newAssignment(c, newDotExpr(c, env.fn).newCall(c))
+    ]
```
alaviss, line 479: Should we mixin the tramp instead?

cps/environment.nim
disruptek: Just so it'll work; this is very old code now. :wink:

cps/environment.nim
disruptek: I thought about that, but there's more reason now to remove the trampoline since...
```nim
trampoline whelp goats()
```
...is so unergonomic.  Also, Zevv would have to move this stuff out of `cps.nim` which kinda defeats the purpose of that file.

cps.nim
disruptek: @Clyybber may even comment on this thread. :shrug:

cps/environment.nim
alaviss: We can just have the tramp so it could be mixed into these create-then-run procs. Having the tramp doesn't mean it has to be used by the user.

Alternatively we can have the dispatcher override the behavior of how this should work outside cps.

disruptek: Yeah, I think creating a hook for this makes a lot of sense.  One question, though, is how to specify such an animal and whether it can/should be specified for a single `.cps.` procedure or for the base continuation type.

I kinda like the idea of overloading the trampoline for an individual continuation type, but maybe that's just gratuitous and the right move is to make the user extend the base type themselves and then they get all the type-based semantic overloads "for free".

PR #124, opened 2021-05-24
cps/xfrm: cleanup exception rewrite and fix bugs

alaviss:
Fixed one main issue where the "current" exception is not refreshed
every bounce and that the exception is not voided after completion
of the except clauses.

Also moved the except rewrite outside of `cpsTryExcept`.

Need more tests.

### DIFF
--- a/cps/xfrm.nim
+++ b/cps/xfrm.nim
@@ -362,20 +362,29 @@ proc rewriteExcept(cont, ex, n: NimNode): tuple[cont, excpt: NimNode] =
     proc consumeException(n: NimNode): NimNode =
       ## Prepend all cpsPending with `setCurrentException nil`
       # XXX: need a way to prevent multiple insertion of this
-      if n.isCpsPending:
+      if n.isScopeExit:
         result = newStmtList()
         # set the global exception to nil
         result.add newCall(bindSym"setCurrentException", newNilLit())
         # set our exception symbol to nil too
         result.add newAssignment(ex.resym(cont, contSym), newNilLit())
         result.add n
+      elif n.kind == nnkReturnStmt:
+        # we are moving to the next continuation
+        result = newStmtList()
+        # set the global exception to nil to prevent leaking it outside
+        result.add newCall(bindSym"setCurrentException", newNilLit())
+        result.add n
 
     result = result.filter(consumeException)
 
   # move the exception body into a handler
-  result.cont = makeContProc(genSym(nskProc, "except"), cont):
-    # rewrite the exception body to be continuation-aware
-    setException(cont, body)
+  result.cont = makeContProc(genSym(nskProc, "except"), cont, body)
+  # rewrite the body of the handler to be exception-aware
+  #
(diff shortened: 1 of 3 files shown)


cps/xfrm.nim
```diff
@@ -322,6 +322,70 @@ macro cpsWhile(cont, cond, n: typed): untyped =
 
   #debug("cpsWhile", result, Transformed, cond)
 
+proc rewriteExcept(cont, ex, n: NimNode): tuple[cont, excpt: NimNode] =
+  ## Rewrite the exception branch `n` to use `ex` as the "current" exception
+  ## and move its body into a continuation
+  doAssert n.kind == nnkExceptBranch
```
disruptek, line 328: What is the point of this line?  Seriously.  :laughing:

cps/xfrm.nim
```diff
@@ -322,6 +322,70 @@ macro cpsWhile(cont, cond, n: typed): untyped =
...
+    else:
+      # add `Type`
+      result.excpt.add n[0]
+
+  proc setException(contSym, n: NimNode): NimNode =
+    ## Set the exception to `ex` before running any other code
+    proc setContException(n: NimNode): NimNode =
+      ## If `n` is a continuation, set the exception to `ex` before
+      ## running any other code
+      # XXX: need a way to prevent multiple insertions of this
```
disruptek, line 351: We could annotate the composed AST with a `discard "exception wuz fixerated"` and then have a check for that, but a better solution would be @saem's NimNode types.

disruptek (approved):
Why not merge?

alaviss:
There are a few major issues left:

- Exception raised and handled within the exception handler will replace the handler's current exception
- Even with the latest commit, the exception might leak if an unhandled exception is raised (we may have to decide how unhandled exceptions affect a running continuation)

disruptek:
> There are a few major issues left:
> 
> * Exception raised and handled within the exception handler will replace the handler's current exception

Okay, but isn't that correct behavior?

> * Even with the latest commit, the exception might leak if an unhandled exception is raised (we may have to decide how unhandled exceptions affect a running continuation)

I don't know why we'd want to catch unhandled exceptions.  Ideally, exceptions don't touch the state so that the user can simply invoke the continuation again -- kinda like quirky exceptions, right?

alaviss:
> Okay, but isn't that correct behavior?

https://play.nim-lang.org/#ix=3nFm <-- apparently not

> I don't know why we'd want to catch unhandled exceptions. Ideally, exceptions don't touch the state so that the user can simply invoke the continuation again -- kinda like quirky exceptions, right?

I worry about this: https://github.com/nim-lang/Nim/issues/18070

disruptek:
> https://play.nim-lang.org/#ix=3nFm <-- apparently not

I guess we have to store all exceptions in the locals.  This is feeling the original handwritten transform, right?

> I worry about this: https://github.com/nim-lang/Nim/issues/18070

Looks like the pointer isn't cleared but that should be an easy fix.  Did you look at the codegen?

PR #126, opened 2021-05-25
cps/xfrm: add a pass to rewrite cps exprs

alaviss:
Flatten expressions with cps calls in it so that it can be rewritten.

Very experimental, need a lot of cleanup and revision to the approach.

### DIFF
--- a/cps/spec.nim
+++ b/cps/spec.nim
@@ -681,3 +681,56 @@ proc isExpr*(n: NimNode): bool =
           return true
     else:
       result = false
+
+proc assignExpr*(n, sym: NimNode): NimNode =
+  ## rewrite the expression `n` into a statement assigning to `sym`
+  ##
+  ## does nothing if `n` is not an expression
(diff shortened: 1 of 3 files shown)


disruptek:
What about putting it in its own file?

alaviss:
> What about putting it in its own file?

You got a name for it? Also I'm using a fair amount of logic already existing in `xfrm`.

disruptek:
Should we break up `xfrm` first, then?  Or simply move more of it elsewhere to make room?

We can do all the return logic in a separate file, for example.

alaviss:
We can't correctly rewrite a call without knowing about `nnkHiddenAddr`:

```nim
var i = 0

proc print(a: var int, b: int) =
  echo a, ' ', b

print i, (noop(); inc i; i)
```

If we are to stash `i` before we pass it to `print`, `print` will not get the correct variable to mutate...

PR #128, opened 2021-05-26
cps/xfrm: make firstReturn aware of more returning control

alaviss:
- Added detection for CPS control-flow annotations
- Added traversal to control-flow blocks like `try`, `finally`, `block`
- Added detection for raise

This should help us reduce the insertion of CPS control-flow where a
returning control will happen before it


disruptek:
Very good.  👍

PR #142, opened 2021-05-30
remove rewrites.cloneProc and use ProcDef fields

saem:
- removes cloneProc usage by introduce ProcDef in more places
- introduce callingParams iterator, for passed in parameter defintions
- create desym operation for ProcDef, which also maintains the invariant
- use returnParam field more pervasively
- setting the returnParam now guarantees the normalized invariant


saem:
All tests and examples passed on my machine.

PR #144, opened 2021-05-31
Typing identdefs

saem:


cps/environment.nim
```diff
@@ -284,7 +264,7 @@ proc newEnv*(c: NimNode; store: var NimNode; via, rs: NimNode): Env=
   store.add check
 
   result = Env(c: c, store: store, via: via, id: via)
-  result.rs = newIdentDefs(ident"result", rs, newEmptyNode())
+  result.rs = newIdentDefs("result", rs)
```
disruptek, line 267: i don't get it.  surely the more elaborately complete and send-documenting version is more ideal, right?

cps/environment.nim
```diff
@@ -297,7 +277,7 @@ proc identity*(e: var Env): NimNode =
 proc initialization(e: Env; kind: NimNodeKind;
                     field: NimNode; value: NimNode): NimNode =
   ## produce the `x = 34` appropriate given the field and identDefs
-  assert kind in {nnkVarSection, nnkLetSection, nnkIdentDefs}
+  doAssert kind in {nnkVarSection, nnkLetSection, nnkIdentDefs}
```
disruptek, line 280: just remove it; we intend to solve these problems with proper typing.

cps/environment.nim
```diff
@@ -312,32 +292,29 @@ proc initialization(e: Env; kind: NimNodeKind;
...
 
-proc letOrVar(n: NimNode): NimNodeKind =
+proc letOrVar(n: IdentDefs): NimNodeKind =
   ## choose between let or var for proc parameters
-  assert n.kind == nnkIdentDefs
-  if len(n) == 2:
-    # ident: type; we'll add a default for numbering reasons
-    n.add newEmptyNode()
-  case n[^2].kind
+  case n.typ.kind:
```
disruptek, line 297: now we're talkin'.

cps/environment.nim
```diff
@@ -470,8 +447,7 @@ proc getException*(e: var Env): NimNode =
   if e.ex.isNil:
     e.ex = genField"ex"
     e = e.set e.ex:
-      nnkVarSection.newTree:
-        newIdentDefs(e.ex, nnkRefTy.newTree(bindSym"Exception"), newNilLit())
+      newVarSection(e.ex, nnkRefTy.newTree(bindSym"Exception"), newNilLit())
```
disruptek, line 450: ah, i see what you did there.  clever.

cps/environment.nim
```diff
@@ -512,12 +487,12 @@ proc createBootstrap*(env: Env; n: ProcDef, goto: NimNode): ProcDef =
...
 
   # do an easy static check, and then
-  if env.rs[1] != result.returnParam:
+  if env.rs.typ != result.returnParam:
     result.body.add:
       result.errorAst:
         "environment return-type doesn't match bootstrap return-type"
   # if the bootstrap has a return type,
-  elif not env.rs[1].isEmpty:
+  elif env.rs.hasType:
```
disruptek, line 495: hello again, my friend.

cps/normalizedast.nim
```diff
@@ -15,6 +15,10 @@ type
 
   ProcDef* = distinct NormalizedNimNode
     ## an nnkProcDef node which has been normalized
+  
+  IdentDefs* = distinct NormalizedNimNode
+
+  VarSection* = distinct NormalizedNimNode
```
disruptek, line 21: omg you may have nailed it.

cps/normalizedast.nim
```diff
@@ -33,6 +37,53 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
   ## desym all occurences of a specific sym
   n.replace(proc(it: NimNode): bool = it == sym, desym sym).NormalizedNimNode
 
+# fn-IdentDefs
+
+nimNodeConverter(IdentDefs)
+
+proc expectIdentDefs*(n: NimNode): IdentDefs =
+  ## return an IdentDef or error out
+  doAssert n.kind == nnkIdentDefs, "not and IdentDefs, got: " & $n.kind
```
disruptek, line 46: not an IdentDefs

cps/normalizedast.nim
```diff
@@ -33,6 +37,53 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
   ## desym all occurences of a specific sym
   n.replace(proc(it: NimNode): bool = it == sym, desym sym).NormalizedNimNode
 
+# fn-IdentDefs
+
+nimNodeConverter(IdentDefs)
```
disruptek, line 42: but really it's a converterToNimNode, right?

cps/normalizedast.nim
```diff
@@ -33,6 +37,53 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
...
+  doAssert n.kind == nnkIdentDefs, "not and IdentDefs, got: " & $n.kind
+  if n[0].kind notin {nnkIdent, nnkSym}:
+    error "bad rewrite presented:\n" & repr(n), n
+  elif n.len != 3:
+    error "bad rewrite, failed to set init\n" & repr(n), n
+
+  return n.IdentDefs
+
+proc newIdentDefs*(n: string, info: NimNode, val = newEmptyNode()): IdentDefs =
+  newIdentDefs(ident(n), info, val).IdentDefs
```
disruptek, line 55: ah, good.

cps/normalizedast.nim
```diff
@@ -33,6 +37,53 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
...
+  n[0]
+
+func typ*(n: IdentDefs): NimNode =
+  n[1]
+
+func val*(n: IdentDefs): NimNode =
+  n[2]
+
+func hasVal*(n: IdentDefs): bool =
+  n.val.kind != nnkEmpty
```
disruptek, line 67: maybe hasValue since it's also hasType.

cps/normalizedast.nim
```diff
@@ -33,6 +37,53 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
...
+
+func inferTypFromImpl*(n: IdentDefs): NimNode =
+  ## returns the typ if specified or uses `macro.getTypeImpl` to infer it
+  if n.hasType: n.typ else: getTypeImpl(n.val)
+
+# fn-VarSection
+
+nimNodeConverter(VarSection)
+
+proc newVarSection*(i: IdentDefs): VarSection =
```
disruptek, line 80: Yes but no because we want to distinguish between identdefs that are mutable versus those that are immutable.  That information can help us optimize ourselves and also let's us put a spin on the output -- _not just an identdefs, but an immutable one_.

cps/normalizedast.nim
```diff
@@ -62,4 +115,11 @@ iterator callingParams*(n: ProcDef): NimNode =
   for a in n.params[1..^1].items:
     yield a
 
-proc desym*(n: ProcDef, sym: NimNode): ProcDef {.borrow.}
\ No newline at end of file
+proc desym*(n: ProcDef, sym: NimNode): ProcDef {.borrow.}
+
+proc addPragma*(n: ProcDef, prag: NimNode) {.borrow.}
+
+proc addPragma*(n: ProcDef, prag: NimNode, pragArg: NimNode) =
```
disruptek, line 122: very good.

cps/transform.nim
```diff
@@ -564,8 +564,7 @@ proc cpsTransformProc*(T: NimNode, n: NimNode): NimNode =
   let booty = env.createBootstrap(n, name)
 
   # we store a pointer to the whelp on the bootstrap
-  booty.addPragma:
-    nnkExprColonExpr.newTree(bindSym"cpsBootstrap", whelp.name)
+  booty.addPragma(bindSym"cpsBootstrap", whelp.name)
```
disruptek, line 567: yes.

cps/transform.nim
```diff
@@ -601,8 +600,8 @@ proc cpsTransformProc*(T: NimNode, n: NimNode): NimNode =
   n.body = env.annotate n.body
 
   # run other stages
-  n.addPragma bindSym"cpsFloater"
-  n.addPragma nnkExprColonExpr.newTree(bindSym"cpsResolver", env.identity)
+  n.addPragma(bindSym"cpsFloater")
+  n.addPragma(bindSym"cpsResolver", env.identity)
```
disruptek, line 604: yes.

cps/environment.nim
```diff
@@ -497,8 +473,7 @@ proc createBootstrap*(env: Env; n: ProcDef, goto: NimNode): ProcDef =
   let c = nskVar.genSym"c"
   result.body.add:
     # declare `var c: Cont`
-    nnkVarSection.newTree:
-      newIdentDefs(c, env.root)
+    cVarSectionToNimNode(newVarSection(c, env.root))
```
disruptek, line 476: but why?

cps/environment.nim
disruptek: _narrator: it wasn't._

cps/transform.nim
disruptek: there's a bracket form, too, right?  maybe via openArray?

disruptek (changes requested):
really great job.  except for var/let section.  but if we can solve identdefs this way it is massive.

disruptek:
also, is there an iterator for identdefs already?  and how about tuples?  works?

cps/environment.nim
```diff
@@ -228,42 +229,21 @@ proc set(e: var Env; key: NimNode; val: NimNode): Env =
   when cpsReparent:
     result.seen.incl key.strVal
 
-iterator addIdentDef(e: var Env; kind: NimNodeKind; n: NimNode): Pair =
-  ## add `a, b, c: type = default` to the env;
-  ## yields pairs of field, value as added
+proc set(e: var Env; key: NimNode; val: VarSection): Env =
+  set(e, key, cVarSectionToNimNode(val))
```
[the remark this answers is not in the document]
saem: If I don't put it here, it'll keep calling itself over and over again and crap out the transform.

cps/environment.nim
```diff
@@ -79,15 +79,16 @@ proc maybeConvertToRoot*(e: Env; locals: NimNode): NimNode =
...
 proc init(e: var Env) =
   if e.fn.isNil:
     e.fn = ident"fn"
   if e.mom.isNil:
     e.mom = ident"mom" # FIXME: use a getter/setter?
   e.id = genSym(nskType, "env")
-  if not e.rs[1].isEmpty:
-    e = e.set(e.rs[0], nnkVarSection.newTree e.rs)
+  if e.rs.hasType:
+    e = e.set(e.rs.name, newVarSection e.rs)
```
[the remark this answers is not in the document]
saem: yeah, that part could be better, need a fancier type to unify all this stuff.

cps/environment.nim
```diff
@@ -79,15 +79,16 @@ proc maybeConvertToRoot*(e: Env; locals: NimNode): NimNode =
     locals
 
 proc set(e: var Env; key: NimNode; val: NimNode): Env
+proc set(e: var Env; key: NimNode; val: VarSection): Env
```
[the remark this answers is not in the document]
saem: I want to get to key typed but it'll take a few successive transforms to get to the right signature.

cps/environment.nim
saem: * `newIdentDefs("result", info = rs)` sucks less?
* `newResultIdentDefs(info = rs)`, sucks less ...

Wait, did that last example just pointed out that this isn't an `IdentDefs` because it's not normalized (no type and no val)?

cps/environment.nim
saem: I need this temporarily, but yeah it needs to go longer term.

This is how I figured out the lesson about all the various forms of `IdentDefs` and how I probably will have to evolve the types a bunch more.

cps/environment.nim
saem: This was the other thing making me suspect I don't think we really want to be making straight `IdentDefs`, we likely want to emit VarSection, LetSection, ProcDefArg, GenericParamDef, etc... or at least have constructors that do the whole production cleanly without denormalization concerns outside of that.

cps/normalizedast.nim
saem: I'd burn this PR just on this alone.

cps/normalizedast.nim
saem: No, I hate it, they're wrong. I'm only tolerating them because the `.NimNode` spam would be high, but as things become more typed we should throw them out ASAP and put in the spam in the few places where it ends up being a forcing function for the type conversion.

cps/normalizedast.nim
saem: Daaaayum, that's a whole different level, so you're thinking:
* pre-normalization <-- we don't type this shit
* post-normalization <-- what we're typing and is mutable
* post-transform <-- this is what you're talking about and it's immutable

Did I get that right?

cps/normalizedast.nim
saem: yeah, I was flip flopping on that one, `hasValue` it is.

Yes, proc, you ARE a good proc. You have value, hasValue, in fact.

cps/environment.nim
saem: it calls the wrong add if I don't force it... because "converters" work so well. 🙄

saem:
There is not an iterator for identdefs, but that's a bit of a can of worms:
* in what situations do you want to iterate over identdefs
* are we talking about procdef params, generic param defs, var/let section runs?
* or in some other general sense?
* because normalizingRewrites is a lot better/regular, reasons to iterate over identdefs are dying I think

Tuples and LetSections can be next on the hit list, that should collapse a bunch more environment code.

cps/transform.nim
saem: Good idea, I'll add a reminder and if we need it then we can pop that in.

cps/environment.nim
disruptek: no, i came around on this one.

cps/normalizedast.nim
disruptek: How is it a forcing function?  It's forcing me to write more than I need to.  If you're a person that buys NimNodes and I offer to sell you an IdentDefs, you are getting more than you paid for, clearly.  So why shouldn't that converter exist?  It makes money every time it's used!

cps/normalizedast.nim
disruptek: You're right that post-transform it's immutable, yes.  What we structure in a var section to signal mutability is:
- var sections
- proc params that are vars
Everything else is immutable:
- let sections
- proc params that are not vars

cps/normalizedast.nim
saem: Because you shouldn't need to sell identdefs, rather we want to get rid on the NimNode buyers as much as possible.

So by forcing us to write it out in the meantime the ugliness encourages typing harder to make the ugly go away.

If we think about why we need say a procdef to NimNode converter. Only reason it should exist is to add in a transformed procdef to the module, float a procdef, or add an inner proc. That's what one or two call sites at most? Whereas we don't need all the macros helpers which are invariant breaking "features" waiting to happen.

cps/environment.nim
saem: I know, but maybe babes is hiding here oh well, future us problem.

cps/normalizedast.nim
saem: Ok so this is for later me thinks.

cps/normalizedast.nim
```diff
@@ -15,27 +15,82 @@ type
...
+  ## return an IdentDef or error out
+  doAssert n.kind == nnkIdentDefs, "not an IdentDefs, got: " & $n.kind
+  if n[0].kind notin {nnkIdent, nnkSym}:
+    error "bad rewrite presented:\n" & repr(n), n
+  elif n.len != 3:
+    error "bad rewrite, failed to set init\n" & repr(n), n
+
+  return n.IdentDefs
+
+proc newIdentDefs*(n: string, info: NimNode, val = newEmptyNode()): IdentDefs =
```
alaviss, line 54:
```suggestion
proc newIdentDefs*(n: string, typ: NimNode, val = newEmptyNode()): IdentDefs =
```

We typically use `info` for lineinfo, not types

alaviss (commented):
looks fine otherwise, great work @saem

PR #152, opened 2021-06-03
head and tail hooks

disruptek:
These hooks run after bootstrapping a continuation; the `head` runs when there is no parent; otherwise, we pass the parent to `tail`.


zevv:
Nice, works for me.

It seems pass() has the arguments in wrong order though

disruptek:
> Nice, works for me.
> 
> It seems pass() has the arguments in wrong order though

This is fixed, though it adds an assignment by default.  Should be cheap and optimized out, most likely.

PR #154, opened 2021-06-05
switching to a continuation base type

disruptek:


disruptek:
This is ready for review but not clean enough to merge.

PR #157, opened 2021-06-05
start introducing normalized let section

saem:


cps/environment.nim
```diff
@@ -33,6 +33,10 @@ type
     ex: NimNode                     # the sym we use for current exception
     mom: NimNode                    # the sym we use for parent continuation
 
+  CachePair* = tuple
```
saem, line 36: I've removed pair, just waiting for a clean CI run before I push.

cps/normalizedast.nim
```diff
@@ -18,7 +18,30 @@ type
...
-  VarSection* = distinct NormalizedNimNode
+  VarLet* = distinct NormalizedNimNode
+  VarLetDef = distinct NormalizedNimNode
+    ## identdef or tuple
+
+  TupleVarLet* = distinct VarLet
+  IdentDefVarLet* = distinct VarLet
+
+  LetSection* = distinct VarLet
+  IdentDefLet* = distinct LetSection
```
saem, line 29: removing this and some related funcs/procs.

cps/environment.nim
```diff
@@ -90,43 +91,31 @@ proc init(e: var Env) =
...
-iterator pairs(e: Env): Pair =
+iterator pairs(e: Env): CachePair =
   var seen = initHashSet[string]()
   for pair in e.allPairs:
     # make sure we're actually measuring gensyms for collision
-    if not seen.containsOrIncl definedName(pair.val).strVal:
+    let name = ident(repr(pair.val.name))
+      ## create an identifier from an typesection/identDef as cached;
+      ## this is a copy and it is repr'd to ensure gensym compat...
+    if not seen.containsOrIncl name.strval:
```
disruptek, line 111: The `pair.val.name` is already an ident/symbol, and could be `val.name` if we cared.  Then you convert it to a string and then an ident, before converting it back to a string.  Why not use `$val.name` instead and call it good?  I don't think we /want/ the genSym'd name in `seen`.

cps/environment.nim
```diff
@@ -231,23 +217,21 @@ proc set(e: var Env; key: NimNode; val: NimNode): Env =
...
 
   let
     field = genField def.name.strVal
-    value = newTree(kind,     # ident: <no var> type = default
-                    newIdentDefs(def.name, stripVar(def.typ), def.val))
+    value = newVarLetIdentDef(
+              kind,
+              def.name,
+              stripVar(def.typ), # ident: <no var> type = default
+              def.val)
```
disruptek, line 232: too much whitespace :laughing:

cps/environment.nim
```diff
@@ -306,20 +292,21 @@ proc letOrVar(n: IdentDefs): NimNodeKind =
...
+    section =
+      if kind in {nnkVarSection, nnkLetSection}:
+        kind
+      else:
+        letOrVar(d)
+    (field, value) = e.addIdentDef(section, d)
+  when cpsDebug == "Env":
+    echo $kind, "\t", repr(d)
+  # XXX: remove the `.NimNode`
+  return e.initialization(kind, field, value.NimNode)
```
disruptek, line 309:
```suggestion
  result = e.initialization(kind, field, value.NimNode)
```

cps/environment.nim
```diff
@@ -351,40 +338,45 @@ when false:
...
-            for field, value in e.addIdentDef(n.kind, expectIdentDefs(entry)):
-              tups.add newDotExpr(child, field)
-          maybeAdd newAssignment(tups, defs.last)
-        else:
-          # an iterator handles `var a, b, c = 3` appropriately
-          for assignment in e.addAssignment(n.kind, expectIdentDefs(defs)):
-            maybeAdd assignment
+    # XXX: this branch goes away once we type procParams, as that's the only
+    #      other use for this proc based on the call sites.
+    doAssert false, "this is a deprecated path and should not be triggered"
```
disruptek, line 375:
```suggestion
    error "this is a deprecated path and should not be triggered"
```

cps/environment.nim
saem: This is following what happened in the preivous `definedName`, so I don't have strong opinions, I'm pretty sure we do want the genSym'd stuff because these feed into the locals?

cps/normalizedast.nim
```diff
@@ -37,57 +59,165 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
...
+
+func inferTypFromImpl*(n: DefLike): NimNode =
+  ## returns the typ if specified or uses `macro.getTypeImpl` to infer it
+  if n.hasType: n.typ else: getTypeImpl(n.val)
+
 # fn-IdentDefs
 
 defineToNimNodeConverter(IdentDefs)
 
+proc validateIdentDefs(n: NimNode): void =
```
alaviss, line 83: No `void` return type please

cps/normalizedast.nim
```diff
@@ -37,57 +59,165 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
...
+  ## validators only, afterwards it's safe to cast, allows re-use
+  if n.kind != nnkIdentDefs:
+    errorGot "not an IdentDefs", n, $n.kind
+  elif n[0].kind notin {nnkIdent, nnkSym}:
+    errorGot "bad rewrite presented", n
+  elif n.len != 3:
+    errorGot "bad rewrite, failed to set init", n
 proc expectIdentDefs*(n: NimNode): IdentDefs =
   ## return an IdentDef or error out
```
alaviss, line 92: The name doesn't signify that it would return an `IdentDefs` to me

cps/normalizedast.nim
```diff
@@ -37,57 +59,165 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
...
+  if n.hasType: n.typ else: getTypeImpl(n.val)
 
-func typ*(n: IdentDefs): NimNode =
-  n[1]
+# fn-VarLet
 
-func val*(n: IdentDefs): NimNode =
-  n[2]
+proc expectVarLet*(n: NimNode): VarLet =
+  ## return a VarLet if this is a var or let section, otherwise error out
```
alaviss, line 162: ditto

cps/normalizedast.nim
```diff
@@ -18,7 +18,29 @@ type
...
+  VarLetDef = distinct NormalizedNimNode
+    ## identdef or tuple
+
+  TupleVarLet* = distinct VarLet
+  IdentDefVarLet* = distinct VarLet
+
+  LetSection* = distinct VarLet
+
+  VarSection* = distinct VarLet
+  IdentDefVar* = distinct IdentDefVarLet
```
alaviss, line 31: None of these have any comments on what they are

cps/normalizedast.nim
```diff
@@ -18,7 +18,29 @@ type
...
+
+  DefLike = IdentDefs | VarLetDef
+
+  LetSectionLike = LetSection
+  VarSectionLike = VarSection | IdentDefVar
+  VarLetLike = VarLet | TupleVarLet | IdentDefVarLet | LetSectionLike |
+               VarSectionLike
+  VarLetIdentDefLike = IdentDefVarLet | IdentDefVar
```
alaviss, line 39: You're using these in public API but you are not exporting them or documenting them.

cps/normalizedast.nim
```diff
@@ -37,57 +59,165 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
   ## desym all occurences of a specific sym
   n.replace(proc(it: NimNode): bool = it == sym, desym sym).NormalizedNimNode
 
+# fn-DefLike
+
+func typ*(n: DefLike): NimNode = n.NimNode[^2]
+func val*(n: DefLike): NimNode = n.NimNode[^1]
```
alaviss, line 65: Use `value` since you already have a `hasValue`

cps/normalizedast.nim
```diff
@@ -37,57 +59,165 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
...
+  ## Yes, proc, you ARE a good proc. You have value, hasValue, in fact.
+  n.val.kind != nnkEmpty
+
+func hasType*(n: DefLike): bool =
+  ## has a non-Empty type (`typ`) defined
+  n.typ.kind != nnkEmpty
+
+func inferTypFromImpl*(n: DefLike): NimNode =
+  ## returns the typ if specified or uses `macro.getTypeImpl` to infer it
+  if n.hasType: n.typ else: getTypeImpl(n.val)
```
alaviss, line 77: This is... confusing. What exactly is a normalized `DefLike`? Is a `DefLike` something that has all 3 nodes populated or is it something else? Normalized should be one thing and one thing only.

cps/normalizedast.nim
```diff
@@ -37,57 +59,165 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
...
 
-func hasType*(n: IdentDefs): bool =
-  ## has a non-Empty type (`typ`) defined
-  n.typ.kind != nnkEmpty
+proc asVarLetTuple*(n: VarLet): TupleVarLet =
+  ## return a TupleVarLet if the def is a VarTuple, otherwise error out
+  validateAndCoerce(n.NimNode, TupleVarLet)
+proc asVarLetIdentDef*(n: VarLet): IdentDefVarLet =
+  ## return a IdentDefVarLet if the def is an IdentDef, otherwise error out
+  validateAndCoerce(n.NimNode, IdentDefVarLet)
```
alaviss, line 173: And we are using `as` here?

cps/normalizedast.nim
```diff
@@ -37,57 +59,165 @@ proc desym*(n: NormalizedNimNode, sym: NimNode): NormalizedNimNode =
...
 
-func inferTypFromImpl*(n: IdentDefs): NimNode =
-  ## returns the typ if specified or uses `macro.getTypeImpl` to infer it
-  if n.hasType: n.typ else: getTypeImpl(n.val)
+# fn-TupleVarLet
+
+proc typ*(n: TupleVarLet): NimNode =
+  ## return the type based on `getTypeInst`
```
alaviss, line 178: Right so what should we expect? Is it a tuple type?

cps/normalizedast.nim
```diff
@@ -18,7 +18,29 @@ type
   
   IdentDefs* = distinct NormalizedNimNode
 
-  VarSection* = distinct NormalizedNimNode
+  VarLet* = distinct NormalizedNimNode
+  VarLetDef = distinct NormalizedNimNode
+    ## identdef or tuple
+
+  TupleVarLet* = distinct VarLet
```
alaviss, line 25: Do you think tuple would work better as a more general type?

```nim
(a, b) = (0, 1)
```

Is a real thing and `var`/`let` is not required to unpack it

cps/normalizedast.nim
```diff
@@ -18,7 +18,29 @@ type
   
   IdentDefs* = distinct NormalizedNimNode
 
-  VarSection* = distinct NormalizedNimNode
+  VarLet* = distinct NormalizedNimNode
+  VarLetDef = distinct NormalizedNimNode
+    ## identdef or tuple
```
alaviss, line 23: More documentation needed, what are these?

alaviss (commented):
Sorry for the late review, I didn't read this carefully when I gave you the pass.

cps/normalizedast.nim
disruptek: Maybe `consumeIdentDefs` makes more sense to you?  I don't feel strongly but maybe you can characterize the action a little better.

cps/normalizedast.nim
disruptek: Okay, but is that bad?

disruptek: _narrator: it was very, **very** bad indeed_

cps/normalizedast.nim
saem: added docs in this pr: https://github.com/disruptek/cps/pull/161/files

cps/normalizedast.nim
saem: I don't think that's particularly interesting as a type, as in when would we actually use it?

The guarantees of normalization are not on tuples, but on nnkVarTuple inside a nnkLetSection or nnkVarSection.

cps/normalizedast.nim
saem: It's some sort of tupletype, but I don't want to add too many types all at once.

The literal translation of every nnkFoo kind having a normalized equivalent isn't actually all that useful. The guarntees (real benefit of types) are on fragments of an AST. So normalize a var section means a var section with a single identdefs or vartuple, but if I give you a identdefs without the context of a var or let section, the guarantee isn't the same.

cps/normalizedast.nim
saem: Do you prefer one over the other or something else?

cps/normalizedast.nim
saem: yeah, I started with this because it errors out, felt righter, less sure about that now.

cps/normalizedast.nim
saem: damnit... fine

cps/normalizedast.nim
saem: fixed here: https://github.com/disruptek/cps/pull/161/files

cps/normalizedast.nim
saem: https://github.com/disruptek/cps/pull/161/files

cps/normalizedast.nim
saem: but it's all symmetric with `typ`. :(

cps/normalizedast.nim
saem: fiiiiiiiiiine

cps/normalizedast.nim
saem: maaaaaybe, keep the suggestions coming, but maybe in IRC or some shit.

cps/normalizedast.nim
alaviss: I would prefer that we use one naming scheme, either `expect` or `as`.

PR #161, opened 2021-06-07
More Typed AST

saem:


cps/environment.nim
```diff
@@ -65,35 +66,36 @@ func isDirty*(e: Env): bool =
...
 
-proc castToRoot(e: Env; n: NimNode): NimNode =
-  newTree(nnkCall, e.root, n)
+proc castToRoot(e: Env; n: NormalizedNode): NormalizedNode =
+  newCall(e.root, n)
 
-proc castToChild(e: Env; n: NimNode): NimNode =
-  newTree(nnkCall, e.identity, n)
+proc castToChild(e: Env; n: Name): NormalizedNode =
+  # XXX: remove NimNode
```
alaviss, line 79: What `NimNode`?

cps/normalizedast.nim
```diff
@@ -1,66 +1,452 @@
...
+  r.Name
+
+proc desym*(n: Name): Name {.borrow.}
+  ## ensures that `Name` is an `nnkIdent`
+proc resym*(fragment: NormalizedNode, sym, replacement: Name): NormalizedNode {.borrow.}
+  ## replace `sym` in the AST `fragment` with the `replacement`
+
+func typeInst*(n: Name): NimNode =
+  ## gets the type via `getTypeInst`
+  ## XXX: this shouldn't be done on name, as these could be bare idents
```
alaviss, line 374: Then I think this should either not exist, or add a `deprecated` warning to it so we can refactor the parts that use it

cps/normalizedast.nim
```diff
@@ -72,39 +458,61 @@ func hasType*(n: DefLike): bool =
   ## has a non-Empty type (`typ`) defined
   n.typ.kind != nnkEmpty
 
-func inferTypFromImpl*(n: DefLike): NimNode =
+func inferTypFromImpl*(n: DefLike): Name =
```
alaviss, line 461: `TypeExpr`?

cps/normalizedast.nim
```diff
@@ -113,6 +521,23 @@ func hasType(n: VarLetLike): bool = n.def.typ.kind != nnkEmpty
...
+    ## abstract over let or var sections with identdefs within
+
+func identdef*(n: VarLetIdentDefLike): IdentDef = n.def.IdentDef
+  ## retrieve the innner IdentDef
+func name*(n: VarLetIdentDefLike): Name = n.identdef.name
+  ## Name (ident|sym) of the identifer, as we only have a single identdefs it
+  ## will have the name
+func inferTypFromImpl*(n: VarLetIdentDefLike): Name =
+  ## returns the typ if specified or uses `macro.getTypeImpl` to infer it
+  if n.hasType: n.typ.Name else: getTypeImpl(n.val).Name
```
alaviss, line 537: This should be removed. The promise we make in `normalizingRewrites` is that there will always be a type.

cps/normalizedast.nim
```diff
@@ -1,66 +1,452 @@
...
+    nnkVarTy, nnkRefTy, # the weirder ones
+    nnkTupleConstr      # still have tuples in a few areas
+    # nnkEmpty          # excluded as it's only allowed in some cases
+  }
+  ## list of type NimNodeKind
+  ## XXX: this is an incomplete list
 
 func errorGot(msg: string, n: NimNode, got: string = repr(n)) =
   ## useful for error messages
   error msg & ", got:\n" & repr(got), n
```
alaviss, line 146: If this is only used for validation in here, make it raise `ValueError` instead.

cps/normalizedast.nim
```diff
@@ -188,69 +600,146 @@ func clone*(n: VarLet, value: NimNode = nil): VarLet =
...
-converter identDefVarToIdentDefVarLet*(n: IdentDefVar): IdentDefVarLet =
-  # allow downgrading
-  n.IdentDefVarLet
+proc newProcDef*(name: Name, formalParams: openArray[NimNode]): ProcDef =
+  ## create a new proc def with name and formal params (includes return type)
+  ## and an empty body (`nnkStmtList`)
+  newProc(name.Nimnode, formalParams, newStmtList()).ProcDef
 
-# fn-ProcDef
+func asProcDef*(n: NimNode): ProcDef =
```
alaviss, line 674:
```suggestion
func asProcDef*(n: NormalizedNode): ProcDef =
```

cps/rewrites.nim
```diff
@@ -397,8 +434,24 @@ func replace*(n: NimNode, match: Matcher, replacement: NimNode): NimNode =
       copyNimTree replacement
     else:
       nil
-
+
```
alaviss, line 437: trailing spaces

alaviss (commented):
Looks much better that what we have right now.

Can you move `genField` to `normalizedast`?

cps/normalizedast.nim
saem: That's not great for debugging is it?

saem: Is that true for proc params and ident defs inside a var or let section?

The other issue is that they dervive type information differently.

cps/normalizedast.nim
saem: That means a bunch of changes to other code in order for us to have a `NormalizedNode` in the first place.

Most annoying part is in `transform.cpsResolver`

cps/normalizedast.nim
saem: Is that true for proc params and ident defs inside a var or let section?

The other issue is that they dervive type information differently.

disruptek:
Do we just merge it so we can move forward with this enterprise?

saem:
> Do we just merge it so we can move forward with this enterprise?

I'll look at it again this weekend, still feels like it should do more with less.

PR #162, opened 2021-06-08
More correct normalizations for hidden nodes

alaviss:


cps/rewrites.nim
```diff
@@ -181,11 +181,24 @@ proc normalizingRewrites*(n: NimNode): NimNode =
...
+        # its significance.
+        result = normalizingRewrites n[1]
+
+      of nnkHiddenStdConv:
+        if n.len == 2 and n[0].kind == nnkEmpty:
+          result = normalizingRewrites n[1]
+        else:
+          raise newException(Defect,
+            "unexpected conversion form:\n" & treeRepr(n))
+
```
disruptek, line 195: You may as well explain what these are and how you're modifying them and why.

cps/rewrites.nim
```diff
@@ -196,27 +209,22 @@ proc normalizingRewrites*(n: NimNode): NimNode =
...
-                  result.add:
-                    normalizingRewrites converted
+              let unwrapped = normalizingRewrites child
+
+              # if the unwrapped node is an nnkBracket and
+              # n is a node that can take more than 2 arguments
+              # and this node is the very last children
+              if unwrapped.kind == nnkBracket and
+                 n.kind in {nnkCall, nnkCommand} and
+                 index == n.len - 1:
```
disruptek, line 219: Prefer dominating conditionals so it's obvious what's going on here and which cases are not covered.  _very last child_

disruptek (approved):
Just remember that test files are expensive -- they represent 1-5 invocations of the compiler -- so I'd just put this into taste.nim or add it to t145.nim and rename that file to tiny.nim or some shit.

PR #169, opened 2021-06-11
cps/transform: give the original symbol of the proc to booty

alaviss:
This enables the booty to take the place of the original proc,
effectively working around nim-lang/Nim#18235

### DIFF
--- a/cps/transform.nim
+++ b/cps/transform.nim
@@ -538,7 +538,7 @@ proc cpsTransformProc*(T: NimNode, n: NimNode): NimNode =
   ## rewrite the target procedure in Continuation-Passing Style
 
   # keep the original symbol of the proc
-  let prcSym = n[0]
+  let originalProcSym = n[0]
   # make the AST easier for us to consume
   var n = normalizeProcDef n
   # establish a new environment with the supplied continuation type;
@@ -581,7 +581,7 @@ proc cpsTransformProc*(T: NimNode, n: NimNode): NimNode =
   # this is a workaround for nim bugs:
   # https://github.com/nim-lang/Nim/issues/18203 (used hints)
   # https://github.com/nim-lang/Nim/issues/18235 (export leaks)
-  booty.name = prcSym
+  booty.name = originalProcSym
 
   # now we'll reset the name of the new proc
   # XXX: not sure why we need to desym, but we have to or it won't be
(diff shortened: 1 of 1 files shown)


cps/transform.nim
```diff
@@ -574,12 +576,17 @@ proc cpsTransformProc*(T: NimNode, n: NimNode): NimNode =
...
 
+  # give the booty the sym we got from the original, which
+  # make the booty take the place of the original proc
+  booty.name = prcSym
+
   # now we'll reset the name of the new proc
-  n.name = name
+  # XXX: not sure why we need to desym, but we have to or it won't be
+  # matched inside booty
+  n.name = desym(name)
```
disruptek, line 586: It sounds like you _do_ know why we need to desym.  But are you sure we cannot merely desym the references inside the bootstrap?

cps/transform.nim
```diff
@@ -537,6 +537,8 @@ macro cpsFloater(n: typed): untyped =
 proc cpsTransformProc*(T: NimNode, n: NimNode): NimNode =
   ## rewrite the target procedure in Continuation-Passing Style
 
+  # keep the original symbol of the proc
+  let prcSym = n[0]
```
disruptek, line 541: No point in abbreviating; we use it so rarely, let's call it what it is as verbosely as needed to convey its meaning.

cps/transform.nim
```diff
@@ -574,12 +576,17 @@ proc cpsTransformProc*(T: NimNode, n: NimNode): NimNode =
   # like magics, the bootstrap must jump
   booty.addPragma ident"cpsMustJump"
 
+  # give the booty the sym we got from the original, which
+  # make the booty take the place of the original proc
```
disruptek, line 580:
```suggestion
  # causes the bootstrap symbol to adopt the original procedure's symbol;
  # this is a workaround for nim bugs:
  # https://github.com/nim-lang/Nim/issues/18203 (used hints)
  # https://github.com/nim-lang/Nim/issues/18235 (export leaks)
```

cps/transform.nim
alaviss: We can't, genSym has the property of not being in the symbol table, thus you can never refer to one by ident.

disruptek: But by the time the compiler gets the input, the symbol will be defined.  Don't we do this in many places already?

PR #170, opened 2021-06-12
stack unwind in response to exceptions

disruptek:
Not much to do; we just need to set the `ex` field on the `Continuation` and jump via the `unwind` in the event of an exception.

@alaviss needs to weigh in.


alaviss (commented):
How would the exception propangate from the source to previous continuations when I don't see any `try-except` to catch any raised exception in the body?

And this doesn't seem to let previous continuations handle the exception unless they're the head?

PR #171, opened 2021-06-13
Exceptions refactoring and try-finally transform

alaviss:
As usual, the compiler blocks our transform: ~~https://github.com/nim-lang/Nim/issues/18254~~ Worked around that bug

Blocked by: https://github.com/nim-lang/Nim/pull/18247

So main changes:
- Moved a few things around
- `try-except` is now implemented by merging all `except` branches into one then turn that into a continuation. This should help with implementing an unwinding system for cps.
- `try-finally` is implemented by turning the `finally` leg into a continuation ~~taking a generic parameter pointing to where to go next.~~ then specialize it to each exit legs.

Known issues:
- ~~`try-except` doesn't work with except branches matching multiple exception types: `except ValueError, IOError`.~~
- ~~`try-finally` doesn't handle early returns, but this is fixable by improving `isScopeExit` and making early return an annotation.~~

### DIFF
--- a/cps/transform.nim
+++ b/cps/transform.nim
@@ -404,85 +404,50 @@ macro cpsTryFinally(cont, ex, n: typed): untyped =
     let finallyBody = tryFinally.last.last
 
     # Turn the finally into a continuation leg.
-    var final = makeContProc(nskProc.genSym("Finally"), cont, finallyBody)
+    let final = makeContProc(nskProc.genSym("Finally"), cont, finallyBody)
 
     # A property of `finally` is that it inserts itself in the middle
     # of any scope exit attempt before performing the scope exit.
     #
-    # To implement this we will turn the finally continuation leg into
-    # a generic that specify where to jump to after the leg finishes.
+    # To achieve this, we generates specialized finally legs that
+    # dispatches to each exit points.
     #
-    # Starting with the generic parameter, we will use
-    # `Fn: static ContinuationProc[Continuation]` as our generic param.
-    #
-    # This parameter will specify the function to jump to after this
-    # continuation.
-    let genericDef =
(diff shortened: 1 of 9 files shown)


cps/transform.nim
```diff
@@ -404,100 +404,63 @@ macro cpsTryFinally(cont, ex, n: typed): untyped =
...
+    # XXX: While this generator works perfectly for our usage, it is not
+    # the best way to template. Preferrably we would use generics, which
+    # is blocked by nim-lang/Nim#18254.
+    proc generateContinuation(templ, replace, replacement: NimNode): NimNode =
+      ## Given a continuation template `templ`, replace all `replace` with
+      ## `replacement`, then generate a new set of symbols for all
+      ## continuations within.
+      # The key is the symbol need updating and the value is what to update
+      # it to
+      var replacements: Table[NimNode, NimNode]
```
disruptek, line 424: Rather than a comment, I'd just use a type declaration. :grin:

cps/transform.nim
```diff
@@ -404,100 +404,63 @@ macro cpsTryFinally(cont, ex, n: typed): untyped =
...
+          replacements[n.name] = genSym(n.name.symKind, n.name.strVal)
+          result.name = replacements[n.name]
+          # Rewrite the body to update the references within it
+          result.body = result.body.filter(generator)
+
+        elif n in replacements:
+          result = copyNimTree(replacements[n])
+
+      replacements[replace] = replacement
+      templ.filter(generator)
```
disruptek, line 445: eh do we ever use it like this?  I think we prefer `result = filter(templ, generator)` right?

disruptek (approved):
Really great advancement.  I also think this stuff should go into its own file; it can be imported into `transform`, right?

disruptek (approved):

PR #175, opened 2021-06-17
cps/transform: implement a context management system for exception

alaviss:
Exception handlers in Nim operates on a stack built within the Exception
object. Thus to ensure that continuations don't corrupt the stack
outside of CPS, we generates a manager that configures the current exception
to the continuation's before running the continuation, then restore the
original on return.

In the case of unhandled exceptions, we restore the old exception stack then
raise the unhandled exception to reconnect them.

Known issue:
- ~~Clearing the exception in the continuation is not reimplemented yet~~
- ~~We are still ignoring unhandled exceptions coming from the continuation~~

### DIFF
--- a/cps/transform.nim
+++ b/cps/transform.nim
@@ -304,7 +304,10 @@ proc wrapContinuationWith(n, cont, replace, templ: NimNode): NimNode =
       result = n
       let nextCont = n.getContSym
       result.body = result.body.wrapContinuationWith(nextCont, replace):
-        templ.resym(cont, nextCont)
+        if cont.kind == nnkSym:
+          templ.resym(cont, nextCont)
+        else:
+          templ
 
   result = copyNimTree templ
   # perform body replacement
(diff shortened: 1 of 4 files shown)


disruptek:
Suggestion: use `genAst` for the body code at the end of `withException`.

cps/transform.nim
```diff
@@ -757,6 +737,91 @@ macro cpsFloater(n: typed): untyped =
...
+        # This will be the manager and will replace `n`
+        result = n.ProcDef.clone(newStmtList())
+        # Obtain the continuation name to replace it
+        result.name = n.name
+        # Take its pragmas too, but without cpsHasException, of course
+        result.pragma = n.pragma.stripPragma("cpsHasException")
+
+        # Desym the continuation symbol to detach it from the inner
+        # continuation
+        result.params[1][0] = desym result.params[1][0]
```
disruptek, line 776: We still can't do this without indexing?  :roll_eyes:

cps/transform.nim
```diff
@@ -757,6 +737,91 @@ macro cpsFloater(n: typed): untyped =
...
+        result.body.add inner.NimNode
+
+        let
+          cont = result.getContSym
+          innerFn = inner.name
+
+          exCont = hasException[1]
+          ex = hasException[2].resym(exCont, cont)
+
+        {.warning: "Compiler bug walkaround".}
```
disruptek, line 789: still `workaround`.  the idea is that you have to do work that you shouldn't have to do via the more direct route.

cps/transform.nim
```diff
@@ -757,6 +737,91 @@ macro cpsFloater(n: typed): untyped =
...
+    ## Build a manager for every continuation with an exception in `n`.
+    ##
+    ## This manager will:
+    ## - Preserve the exception outside of CPS on enter-exit.
+    ## - Set the exception to what the continuation want for its duration
+    ## - Should the continuation raise, connect the exception to the stack
+    ##   outside of CPS.
+    if n.isCpsCont:
+      let hasException = n.pragma.findChild(it.kind == nnkCall and
+                                            it[0] == bindSym"cpsHasException")
```
disruptek, line 754: if we cannot use `hasPragma`, we need something here that we _can_ use.

cps/transform.nim
```diff
@@ -757,6 +737,91 @@ macro cpsFloater(n: typed): untyped =
...
+      # If `n` is a continuation with exception
+      if hasException != nil:
```
disruptek, line 756: prefer `isNil`

cps/transform.nim
```diff
@@ -757,6 +737,91 @@ macro cpsFloater(n: typed): untyped =
...
+        # Create a copy of this continuation and rename it
+        let inner = n.ProcDef.clone(n.body)
```
disruptek, line 758: prefer `clone(n.ProcDef, body = n.body)`

more of these follow...  we want to emphasize the verb, then the noun.

cps/transform.nim
```diff
@@ -757,6 +737,91 @@ macro cpsFloater(n: typed): untyped =
...
+        # Now let's go down to business
+        # Add the inner continuation inside the manager
+        result.body.add inner.NimNode
+
+        let
+          cont = result.getContSym
+          innerFn = inner.name
+
+          exCont = hasException[1]
+          ex = hasException[2].resym(exCont, cont)
```
disruptek, line 787: Why don't we unpack these when we first establish `hasException` so I can remember what the hell this stuff means.  It needs a comment in any event.

cps/transform.nim
```diff
@@ -757,6 +737,91 @@ macro cpsFloater(n: typed): untyped =
...
+
+        {.warning: "Compiler bug walkaround".}
+        # XXX: Depends too much on `ex` implementation
+        # The resym earlier should've swapped the symbol for us, but somehow
+        # exCont wasn't equal to the continuation symbol in ex, even though
+        # they were equal when it was set by withException.
+        #
+        # We manually swap the symbol here for because of that.
+        if ex[0][1] != cont:
+          ex[0][1] = cont
```
disruptek, line 797: You may as well not bother with the `if`, right?

cps/transform.nim
```diff
@@ -757,6 +737,91 @@ macro cpsFloater(n: typed): untyped =
...
+            except:
+              # If the continuation raise an unhandled exception,
+              # capture it.
+              let e = getCurrentException()
+              # Restore the old exception
+              setCurrentException(oldException)
+              # Now reraise the continuation's exception, which will
+              # make `oldException` as `e`'s parent, preserving the
+              # exception stack outside of CPS
+              raise e
```
disruptek, line 820: What did you think of `genAst` here?

cps/transform.nim
```diff
@@ -356,6 +313,33 @@ proc wrapContinuationWith(n, cont, replace, templ: NimNode): NimNode =
...
+      # Tag the continuation with details on where to get
+      # the exception symbol.
+      #
+      # We need to store both the originating continuation and the
+      # access AST since pragmas are considered to be *outside*
+      # of the procedure and thus will bind to symbols outside if they
+      # are not already bound.
+      result.pragma.add:
+        newCall(bindSym"cpsHasException", cont, ex)
+      # Rewrites the inner continuations too
```
disruptek, line 335:
```suggestion
      # Rewrite the inner continuations, too.
```

cps/transform.nim
```diff
@@ -356,6 +313,33 @@ proc wrapContinuationWith(n, cont, replace, templ: NimNode): NimNode =
...
+      result = result.withException(cont, ex)
+
+  # XXX: This is just here to verify that the issue we saw in cpsManageException
+  # is not a bug in CPS
+  doAssert cont == ex[0][1]
```
disruptek, line 340: Lazy.  Not a fan.  :wink:

cps/transform.nim
```diff
@@ -356,6 +313,33 @@ proc wrapContinuationWith(n, cont, replace, templ: NimNode): NimNode =
...
+  filter(n, marker)
```
disruptek, line 341: Prefer an explicit result assignment this low on the page.

disruptek (changes requested):
I still like the approach; never had a problem with the logic.

cps/transform.nim
alaviss: Need @Clyybber's `getPragma`

cps/transform.nim
alaviss: It works, the capture interface is terrible IMO.

Also there's a bug with emitting result so you must use `result = ident"result"`.

PR #180, opened 2021-06-20
`...` and return values from child continuations

disruptek:


cps/transform.nim
```diff
@@ -640,13 +640,41 @@ proc annotate(parent: var Env; n: NimNode): NimNode =
...
+          let etype = pragmaArgument(assign.val, "cpsEnvironment")
+          let c = nskVar.genSym"child"  # XXX: workaround for genAst?
+          var shim =
+            genAst(c, etype, tipe = assign.typ, store = assign.name,
+                   call = assign.val):
+              var store: tipe
+              block:
+                var c: Continuation = whelp call
+                while c.running:
+                  c = c.fn(c)
```
alaviss, line 661: Can't we just `tramp` it?

Also it looks to me that this will definitely error out when `c` got dismissed, which doesn't sound very reasonable.

alaviss (changes requested):
Need a test where the child is dismissed, and documentation for the breaking change of `whelp` (returns `Env` instead of `Cont`).

cps/transform.nim
disruptek: Why wouldn't it error out?

And, we need to reproduce the trampoline so that `annotate` will convert it into continuation legs; otherwise, it's no longer a continuation!

alaviss: > Why wouldn't it error out?

Because if you call a `foo()` that doesn't return anything, then it can be dismissed however and it doesn't matter, after `foo()` completes it returns to `mom`.

Now if you call a `bar()` that returns something, then suddenly it errors out when `bar()` got dispatched elsewhere.

> And, we need to reproduce the trampoline so that annotate will convert it into continuation legs; otherwise, it's no longer a continuation!

wdym?

cps/environment.nim
```diff
@@ -438,11 +418,32 @@ proc genException*(e: var Env): NimNode =
...
+  let field =
+    if env.rs.hasType:
+      env.get                  # the return value is env.result
+    else:
+      nnkDiscardStmt.newTree:
+        newEmptyNode()         # the return value is void
+
+  result = ProcDef:
+    genAst(field, c = env.first, cont = env.identity, tipe = env.rs.typ):
+      proc `...`(c: cont): tipe {.used.} =
```
alaviss, line 432: Maybe we can always export this. The only way to use it is to have the Env type, which can only be obtained if the originating proc is exported.

alaviss: On an another thought, maybe not, as it will make cps incompatible for use with procs not on top-level.

disruptek: I don't think we can successfully export it in any event, so we're safe from having to worry about it.

PR #182, opened 2021-06-22
cps: implements a pass for flattening expression

alaviss:
To enable CPS to transform expressions like:

```nim
let x =
  if true:
    noop()
    1
  else:
    -1
```

This PR implements a pass that rewrites expressions into statements, the example above would become:

```nim
var :tmp: int
if true:
  noop()
  :tmp = 1
else:
  :tmp = -1

let x = :tmp
```

Allow CPS to transform as normal.

What really happens here is that we encode the control flow into the AST (ie. the `if` has to be run before `x` is declared).

Checklist:
- [x] Rewrite complex expression to assignment
  - [x] Expression list
  - [x] Block expression
  - [x] If expression
  - [x] Case expression
  - [x] Try expression
- [x] Rewrite complex expressions in
  - [x] Let/Var section
  - [x] Assignments
  - [x] While condition
  - [x] If condition
  - [x] Case matching expression
  - [x] Call parameters
  - [x] Conversion (simpler than call since it's side-effect free and can be applied directly to the expression)
  - [x] Array construction
  - [x] Object construction
  - [x] Tuple construction
  - [x] `and`/`or` boolean operators
  - [x] Discard statement
  - [x] Return statement
  - [x] Raise statement

Let me know if I miss any potential places

Closes #126

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -73,7 +73,7 @@ macro cps*(T: typed, n: typed): untyped =
       # Typically we would add these as pragmas, however it appears
       # that the compiler will run through macros in proc pragmas
       # one-by-one without re-seming the body in between...
-      {.warning: "compiler bug workaround".}
+      {.warning: "compiler bug workaround, see: https://github.com/nim-lang/Nim/issues/18349".}
       result =
         # Add the main transform phase
         newCall(bindSym"cpsTransform", T):
--- a/cps.nim
+++ b/cps.nim
@@ -1,5 +1,5 @@
 import std/[macros]
-import cps/[spec, transform, rewrites, hooks]
+import cps/[spec, transform, rewrites, hooks, exprs]
 export Continuation, ContinuationProc
 export cpsCall, cpsMagicCall, cpsVoodooCall, cpsMustJump
 
@@ -68,7 +68,20 @@ macro cps*(T: typed, n: typed): untyped =
   when defined(nimdoc):
     n
   else:
-    cpsTransformProc(T, n)
+    case n.kind
+    of nnkProcDef:
+      # Typically we would add these as pragmas, however it appears
+      # that the compiler will run through macros in proc pragmas
+      # one-by-one without re-seming the body in between...
+      {.warning: "compiler bug workaround".}
+      result =
+        # Add the main transform phase
+        newCall(bindSym"cpsTransform", T):
+          # Add the flattening phase which will be run first
+          newCall(bindSym"cpsFlattenExpr"):
+            n
+    else:
+      result = getAst(cpsTransform(T, n))
 
 proc makeErrorShim(n: NimNode): NimNode =
   ## Upgrades a procedure to serve as a CPS primitive, generating
--- a/cps/exprs.nim
+++ b/cps/exprs.nim
@@ -26,7 +26,7 @@ func hasCpsExpr(n: NormalizedNimNode): bool =
     of nnkVarSection, nnkLetSection:
       let n = expectVarLet n
       result = n.val.NormalizedNimNode.hasCpsExpr
-    of nnkElifBranch, nnkElifExpr:
+    of nnkElifBranch, nnkElifExpr, nnkWhileStmt:
       result = n[0].NormalizedNimNode.hasCpsExpr
     of nnkStmtList, nnkStmtListExpr, nnkIfStmt, nnkIfExpr, nnkCaseStmt:
       for child in n.items:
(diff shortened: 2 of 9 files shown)


cps.nim
```diff
@@ -68,7 +68,20 @@ macro cps*(T: typed, n: typed): untyped =
   when defined(nimdoc):
     n
   else:
-    cpsTransformProc(T, n)
+    case n.kind
+    of nnkProcDef:
+      # Typically we would add these as pragmas, however it appears
+      # that the compiler will run through macros in proc pragmas
+      # one-by-one without re-seming the body in between...
+      {.warning: "compiler bug workaround".}
```
disruptek, line 76: Please provide the issue number or, ideally, the URL, so we can remove it eventually.

cps/returns.nim
```diff
@@ -13,9 +13,9 @@ proc firstReturn*(p: NimNode): NimNode =
       result = child.firstReturn
       if not result.isNil:
         break
-  of nnkBlockStmt, nnkBlockExpr, nnkFinally:
+  of nnkBlockStmt, nnkBlockExpr, nnkFinally, nnkPragmaBlock:
     result = p.last.firstReturn
-  elif p.isCpsPending or p.isCpsBreak or p.isCpsContinue:
+  elif p.isScopeExit:
```
disruptek, line 18: Funny, I just added this to my branch. :+1:

cps/rewrites.nim
```diff
@@ -354,7 +306,75 @@ proc workaroundRewrites*(n: NimNode): NimNode =
           rewriteContainer:
             workaroundRewrites child
 
-  result = filter(n, workaroundSigmatchSkip)
+  proc workaroundCompilerBugs(n: NimNode): NimNode =
```
disruptek, line 309: Good.

cps/exprs.nim
```diff
@@ -0,0 +1,752 @@
...
+      # Add the rewritten body
+      result.add:
+        NimNode:
+          filterExpr(NormalizedNimNode(branch.last), transformer)
+    else:
+      result = NormalizedNimNode:
+        n.errorAst "unexpected node kind in case/if expression"
+
+  case n.kind
+  of AtomicNodes, CallNodes, nnkTupleConstr, nnkObjConstr, nnkConv, nnkBracket:
```
disruptek, line 86: so, `ConstructNodes` and `nnkConv`.

cps/exprs.nim
```diff
@@ -0,0 +1,752 @@
...
+    # the part before it because we won't touch them.
+    for idx in 0 ..< n.len - 1:
+      result.add copy(n[idx])
+
+    # Rewrite the last expression to assign to location.
+    result.add:
+      # Convert back to NimNode explicitly because the compiler
+      # can't handle our awesome nodes (our node binds to both
+      # varargs and normal form of add).
+      {.warning: "Compiler workaround here".}
```
disruptek, line 103: But which?  It needs a URL.

cps/exprs.nim
```diff
@@ -0,0 +1,752 @@
...
+    result.add copy(n[0])
+    # Rewrite and add the body
+    result.add:
+      NimNode:
+        filterExpr(NormalizedNimNode(n[1]), transformer)
+  of nnkIfStmt, nnkIfExpr:
+    # It appears that the type of the `if` expression remains if we
+    # don't destroy it by creating a new node instead of copying and
+    # causes all sort of errors.
+    {.warning: "compiler workaround here".}
```
disruptek, line 118: Another.  And, we might have predicted this, right.

cps/exprs.nim
```diff
@@ -0,0 +1,752 @@
...
+    for branch in n.items:
+      result.add:
+        NimNode:
+          rewriteElifOf:
+            NormalizedNimNode branch
+  of nnkCaseStmt:
+    # It appears that the type of the `case` expression remains if we
+    # don't destroy it by creating a new node instead of copying and
+    # causes all sort of errors.
+    {.warning: "compiler workaround here".}
```
disruptek, line 130: Another.

cps/exprs.nim
```diff
@@ -0,0 +1,752 @@
...
+
+func isMutable(n: NormalizedNimNode): bool =
+  ## Determine whether `n` is mutable, as in if `n` value can be mutated by
+  ## changes in the program state.
+  case n.kind
+  of nnkSym:
+    result = n.symKind notin {
+      nskGenericParam, nskParam, nskModule, nskType, nskLet, nskConst, nskProc,
+      nskFunc, nskMethod, nskIterator, nskConverter, nskMacro, nskTemplate,
+      nskEnumField, nskField, nskLabel
```
disruptek, line 204: Surely you can just invert the list with set arithmetic, or put them into a `const` -- or two constants that we check statically to ensure they are comprehensive.

cps/exprs.nim
```diff
@@ -0,0 +1,752 @@
...
+    else:
+      it = assignTo(dst):
+        # debugAnnotation wrap our typed body in a stmtlist, so we take it
+        # out
+        NormalizedNimNode it[0]
+
+macro cpsExprConv(T, n: typed): untyped =
+  ## Apply the conversion to `T` directly into `n`'s trailling expressions.
+  # If we don't shadow this parameter, it will be nnkNilLit.
+  {.warning: "compiler workaround here".}
```
disruptek, line 342: ;-)

disruptek (changes requested):
Looks great.  Let's get this in and start playing with it.

PR #196, opened 2021-06-30
move State and bind it for `()`

disruptek:
This solves a clash in eventqueue between `Event.Finished` and `State.Finished` inside our generated `()`.


disruptek:
cc @riderfighter @willyboar

PR #213, opened 2021-07-18
fixes #211 handle implicit dot expr return value

### LINKED ISSUE #211: Error: cps doesn't know how to rewrite this into assignment:
```
import cps

type
  C = Continuation
  P = object
    val: int

proc bar(c: C): C {.cpsMagic.} =      
  c

proc foo(p: P): int {.cps:C.} =
  bar()             
  p.val
       
echo foo(P())
```

```
/tmp/t.nim(13, 23) template/generic instantiation of `cps` from here
/tmp/t.nim(14, 3) template/generic instantiation of `cpsAsgn` from here
/tmp/t.nim(15, 4) Error: cps doesn't know how to rewrite this into assignment:
DotExpr
  Sym "p"
  Sym "val"

```

saem:


tests/treturns.nim
```diff
@@ -26,14 +26,30 @@ suite "returns and results":
...
+      ## implicit return values via bootstrap
+      type
+        P = object
+          val: int
+
+      var k = newKiller 1
+      proc foo(p: P): int {.cps: Cont.} =
+        noop()
+        step 1
+        p.val
```
alaviss, line 49: The test in `texprs` should be suffice, no?

PR #215, opened 2021-07-22
implements dot expression, addr, deref and magic calls flattening

### LINKED ISSUE #207: data gets lost when using result.add on string result in cps proc
```
import cps

type C = Continuation

proc getchunk(): string {.cps:C.} =
  return "chunk"    

proc getstring(): string {.cps:C.} =
  result.add getchunk()
  echo "1> ", result
  
proc foo() {.cps:C.} = 
  echo "2> ", getstring() 

foo()
```

output:
```
1>
2>
```

expected output:
```
1> chunk
2> chunk
```

### LINKED ISSUE #211: Error: cps doesn't know how to rewrite this into assignment:

alaviss:
Summary:
- Implements `expr().b` flattening
- Implements `expr()[]` flattening
- ~~Error on `addr expr()`~~ Implements `var T` support for `{.magic.}` calls

Supersede and closes #213 

Fixes #211 
Fixes #207

### DIFF
--- a/cps/exprs.nim
+++ b/cps/exprs.nim
@@ -78,9 +78,9 @@ func filterExpr[T: NormNode](n: T, transformer: proc(n: T): T): T =
       result = n.errorAst "unexpected node kind in case/if expression"
 
   case n.kind
-  of AtomicNodes, CallNodes, ConstructNodes, nnkConv, nnkBracketExpr:
-    # For calls, conversions, constructions, constants and basic symbols, we
-    # just emit the assignment.
+  of AccessNodes, CallNodes, ConstructNodes, nnkConv:
+    # For calls, conversions, constructions, constants, basic symbols, field
+    # access, we just emit the assignment.
     result = transformer(n)
   of nnkStmtList, nnkStmtListExpr:
     result = copyNodeAndTransformIt(n):
@@ -162,12 +162,6 @@ func filterExpr[T: NormNode](n: T, transformer: proc(n: T): T): T =
     ## Hidden conversion nodes can be reconstructed by the compiler if needed,
     ## so we just skip them and rewrite the body instead.
     result = filterExpr(n.last, transformer)
-  of nnkDotExpr:
-    ## this is either the first expr we encountered or we were passed this and
-    ## so we treat it as a terminal operation regardless
-    ##
-    ## see https://github.com/disruptek/cps/issues/211).
-    result = transformer(n)
   else:
     result = n.errorAst "cps doesn't know how to rewrite this into assignment"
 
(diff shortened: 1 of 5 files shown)


alaviss:
@disruptek Pretty sure a fail with `refc` can't be a CI pass: https://github.com/disruptek/cps/pull/215/checks#step:5:34

disruptek:
Pretty sure `danger` doesn’t have to pass, but check the code and see.

cps/exprs.nim
```diff
@@ -647,49 +678,97 @@ func annotate(n: NormNode): NormNode =
...
+                # {.magic.} calls, unlike normal procedure, the compiler will not emit `nnkHiddenAddr`
+                # for `var` parameters, thus making regular analysis incorrect.
+                #
+                # As a workaround, we have to obtain this information directly from the symbol definition.
+                if magicType.typeKind == ntyVar and not magicType.sameType(result.typ):
+                  # If the parameter is a `var T` but the type differs from the operand.
+                  # Modify the operand type to `var operand.typ`.
+                  result.typ = TypeExpr nnkVarTy.newTree(result.typ)
+                  # Modify the mutable analysis to check whether the operand location is mutable instead.
+                  result.mutable = child.isMutableLocation
```
saem, line 724: So we force the type on the operand to be mutable (`nnkVarTy` wrapper), but then we check again to see if the original type for the operand is a mutable location. 

Isn't it simply mutable at this point?

alaviss: > Isn't it simply mutable at this point?

Nope.

What we look for is whether the value of the expression can be changed. For example:

```nim
var x: string
add(x, "foo")
#   ^- x is mutable (is a var)
#   However, `x` is not what is being *passed* to `add`.
#   What is being passed is the reference to `x`, which is not mutable (`x` is stored in a static location)
```

In typed AST there will be a `nnkHiddenAddr` wrapping `x` and `isMutable()` knows what to do.

However, in the case of magic calls, this doesn't happen (we would just get `x`), so we have to use the location analysis explicitly.

PR #220, opened 2021-07-24
rewrite return then annotate

### LINKED ISSUE #219: "... has no type" compilation error on `return`
This

```
import cps
   
type C = Continuation

proc foo(n: int): int {.cps:C.} =
  return n * 2

proc bar(n: int): int {.cps:C.} =
  # result = foo(n) works
  return foo(n)  

echo bar(42)
```

Does this:

```
/tmp/t2.nim(9, 25) template/generic instantiation of `cps` from here
/tmp/t2.nim(10, 14) Error: expression '
proc Post Child_402654182(continuation: C): C {.cpsLift(), nimcall, cpsCont().} =
  mixin tail
  mixin head
  mixin unwind
  mixin pass
  mixin dealloc
  mixin alloc
  mixin trace
  mixin coop
  if not isNil(continuation.ex):
    try:
      raise continuation.ex
    finally:
      continuation.ex = nil
  {.cpsPending().}

continuation.fn = Post Child_402654182
cps environment_402654141(continuation).child_402654168 = cps environment_402653389(tail(
    Continuation(continuation),
    C(whelp_402653402(cps environment_402654141(continuation).n_402654142))))
return typeof(continuation) do:
  Continuation(cps environment_402654141(continuation).child_402654168)' has no type (or is ambiguous)
```

Note that changing `return foo(n)` into `result = foo(n)` solves the problem

alaviss:
We are rewriting return to assignment to result, so run annotate after
that so it could cover "assign with cps call"

Fixes #219

### DIFF
--- a/cps/environment.nim
+++ b/cps/environment.nim
@@ -352,7 +352,7 @@ proc localSection*(e: var Env; n: RoutineParam; into: NimNode = nil) {.borrow.}
   ## consume proc definition params and yield name, node pairs representing
   ## assignments to local scope.
 
-proc rewriteReturn*(e: var Env; n: NimNode): NimNode =
+proc rewriteReturn*(e: var Env; n: NormNode): NormNode =
   ## Rewrite a return statement to use our result field.
   if n.len != 1:
     result = n.errorAst "return len != 1"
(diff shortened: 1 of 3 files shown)


disruptek:
When it's this obvious, feel free to just merge and version it, right. :+1:

PR #228, opened 2021-08-08
write the "stack" frames during uncaught exceptions

disruptek:


disruptek:
No opinions?

cps/spec.nim
```diff
@@ -357,6 +349,97 @@ macro cpsMagic*(n: untyped): untyped =
...
+
+proc trampoline*[T: Continuation](c: T): T =
+  ## This is the basic trampoline: it will run the continuation
+  ## until the continuation is no longer in the `Running` state.
+  var c: Continuation = c
+  while not c.isNil and not c.fn.isNil:
+    try:
+      c = c.fn(c)
+    except Exception:
+      writeStackFramesImpl c
```
alaviss, line 421: Why don't we use `writeStackFrames` instead?

alaviss (commented):
LGTM for the most part.

cps/spec.nim
disruptek: It's voodoo or whatever so we can't use it directly.

PR #257, opened 2021-11-07
Update coroutine example, add walktrough

ZoomRmc:
I can't stand behind the correctness of my explanation yet, so please review.

I also suspect it's more of a tutorial material, so separating into its own file could be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,24 +225,24 @@ echo "i counted ", later(), " trips through the goto"
 shows a basic implementation of coroutines communicating with each other on top
 of CPS.
 
-We're going to follow through the execution of this example looking into the 
+We're going to walk through the execution of this example looking into the
 details as we go, but first let's introduce the `Coroutine` type:
 
 ```nim
 type
   Coroutine = ref object of Continuation
     data: int
-    cResume: Coroutine
+    next: Coroutine
 ```
 
-This is a Continuation extended with the `data` on which we're going to perform
-a computation and a Coroutine object which has a pointer to some continuation
-leg. This will allow us to resume the execution of one coroutine from the other.
+This is a `Continuation` extended with the `data` on which we're going to perform
+a computation and a reference to some other `Coroutine` object. This will allow
+us to resume the execution of one coroutine from the other.
 
 The execution starts with instantiating our coroutines, which is necessary for
 continuations. The order is reversed because we need to pass a specific instance
 of the consumer to the filter. `filter` also takes an anonymous function as its
-second argumet, in our case it's a simple doubling written in a short form
+second argument; in our case it's a simple doubling written in a short form
 provided by the `std/sugar` module.
 
 ```nim
@@ -254,12 +254,12 @@ Both coroutines are already "running" (`coroX.running == true`) but nothing is
 executed yet. For this we need to actually launch them:
 
 ```nim
-coro1.resume()
-coro2.resume()
+discard coro1.resume()
+discard coro2.resume()
 ```
 
 The `resume` proc is our dispatcher for the coroutines and it's what actually
-calls the execution via the function pointer inside the continuation:
+runs the continuation via the call to a function pointer held by continuation:
 
 ```nim
 proc resume(c: Coroutine): Coroutine {.discardable.} =
@@ -269,8 +269,13 @@ proc resume(c: Coroutine): Coroutine {.discardable.} =
   result = Coroutine c
 ```
 
-Before invoking that function we need to convert the Coroutine to its base class
-first. Then, the filter coroutine gets launched:
+Actually, this is such a basic pattern for CPS code that the library provides a
+[trampoline](https://nim-works.github.io/cps/cps/spec.html#trampoline%2CsinkT)
+template for this, and it would be preferable to just use it here. Also, notice
+how before invoking the function pointer we needed to convert a `Coroutine` to
+its base type so we could set `c` to `fn(c)`, which returns a `Continuation`.
+
+So, `resume` launches the `filter` coroutine:
 
 ```nim
 proc filter(dest: Coroutine, f: proc(x: int): int) {.cps:Coroutine.} =
@@ -281,24 +286,27 @@ proc filter(dest: Coroutine, f: proc(x: int): int) {.cps:Coroutine.} =
 ```
 
 As the first coroutine launches, it yields (suspends execution) by calling
-`jield` immediately (*yeild* is a keyword in Nim). This is necessary because we
+`jield` immediately (*yield* is a keyword in Nim). This is necessary because we
 haven't actually sent any data to process, yet we need the Coroutines launched
 and waiting for it. We could move the suspension points later, but we don't want
 the coroutines processing the initialized-by-default value prior to the data we
-send them.
+send them (try moving "jields" around and inspect the results).
 
 ```nim
 proc jield(c: Coroutine): Coroutine {.cpsMagic.} =
-  c.cResume = c
+  c.next = c
   return nil
 ```
 
 Our `jield` proc is a special function, as signified by the `{.cpsMagic.}`
 pragma. It allows controlling the execution flow of the continuation by
 returning the continuation leg to run next. Usually, it's the same as the proc's
-argument, but in our case, we store the passed continuation in the `cResume`.
-This way our coroutine could be resumed later, and since the `jield` returns `nil`
-the coroutine gets suspended.
+argument, but in our case, we store the passed continuation in the `next`.
+This way our coroutine could be resumed later.
+
+Since the `cpsMagic` `jield` returns `nil` the coroutine gets suspended. Notice,
+how CPS manages the control flow for us and hides the implementation details
+behind ~~a veil of magic~~ code transformations.
 
 `coro2` is launched next by calling `resume` and in the same manner `consumer`
 starts running and yields immediately.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
 
-### TBD
+### Walkthrough for the `coroutine.nim` example
 
-We'll talk about voodoo here and walk through the coroutine demo, since it pulls together prior concepts and adds voodoo and multiple continuations.
+[examples/coroutine.nim](https://github.com/nim-works/cps/blob/master/examples/coroutine.nim)
+shows a basic implementation of coroutines communicating with each other on top
+of CPS.
+
+We're going to follow through the execution of this example looking into the
```
disruptek, line 228: ... walk through ... yeah.

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+of CPS.
+
+We're going to follow through the execution of this example looking into the 
+details as we go, but first let's introduce the `Coroutine` type:
+
+```nim
+type
+  Coroutine = ref object of Continuation
+    data: int
+    cResume: Coroutine
```
disruptek, line 235: let's just call it anything other than `cResume`.

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+```
+
+This is a Continuation extended with the `data` on which we're going to perform
+a computation and a Coroutine object which has a pointer to some continuation
```
disruptek, line 239: either backquote both symbols, or neither?

`cResume` is a `Coroutine`; ie. a `ref`, and not to a continuation leg, but an entire continuation.

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+leg. This will allow us to resume the execution of one coroutine from the other.
+
+The execution starts with instantiating our coroutines, which is necessary for
+continuations. The order is reversed because we need to pass a specific instance
+of the consumer to the filter. `filter` also takes an anonymous function as its
+second argumet, in our case it's a simple doubling written in a short form
```
disruptek, line 245: _argument_ followed by `;`

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+Both coroutines are already "running" (`coroX.running == true`) but nothing is
+executed yet. For this we need to actually launch them:
+
+```nim
+coro1.resume()
+coro2.resume()
+```
+
+The `resume` proc is our dispatcher for the coroutines and it's what actually
+calls the execution via the function pointer inside the continuation:
```
disruptek, line 262: Do we really want to go into the detail of the function pointer inside the continuation?  I would probably just say, `it's what actually runs the continuation` but if you think it's important, because you want to show `.fn`, that's fine, too.

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+
+```nim
+proc resume(c: Coroutine): Coroutine {.discardable.} =
+  var c = Continuation c
+  while c.running:
+    c = c.fn(c)
+  result = Coroutine c
+```
```
disruptek, line 270: Ugh; `discardable` is poor style, but whatever.  This is where I'd want to use our `trampoline` and, if not, we should explain why we have to do the object conversion.  Otherwise, this is likely to trip someone up when they try to implement it naively.

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+```nim
+proc filter(dest: Coroutine, f: proc(x: int): int) {.cps:Coroutine.} =
+  while true:
+    jield()
+    let n = f(recv())
+    dest.send(n)
+```
+
+As the first coroutine launches, it yields (suspends execution) by calling
+`jield` immediately (*yeild* is a keyword in Nim). This is necessary because we
```
disruptek, line 284: _yield_

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+  c.cResume = c
+  return nil
+```
+
+Our `jield` proc is a special function, as signified by the `{.cpsMagic.}`
+pragma. It allows controlling the execution flow of the continuation by
+returning the continuation leg to run next. Usually, it's the same as the proc's
+argument, but in our case, we store the passed continuation in the `cResume`.
+This way our coroutine could be resumed later, and since the `jield` returns `nil`
+the coroutine gets suspended.
```
disruptek, line 301: Maybe two sentences here, and explain who does the suspending and how that's even a thing.

What's tricky about CPS is remembering the _source_ of control-flow is actually somewhere else, or as we say, "in Nim".  The dispatcher, right.  So we want to get people thinking about how nice it is to be inside CPS where they don't have to give a second thought to how all this works...  but for the purposes of a tutorial, we want to emphasize the difference and highlight the machinery.

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+the `filter` continuation. There we receive the data and try to process it with
+the passed function:
+
+```nim
+let n = f(recv())
+```
+
+The `recv` proc is another special one. `{.cpsVoodoo.}` allows returning the
+contents of the concrete instance of the running continuation (which is not
+accessible directly from the coroutine code of `filter` and `consumer`):
```
disruptek, line 324: Very good.

I would also remark on the ways that `cpsVoodoo` and `cpsMagic` are similar and different.

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+
+```nim
+proc recv(c: Coroutine): int {.cpsVoodoo.} =
+  c.data
+```
+
+After applying the `f` proc to the received data we pass in to the `consumer`
+coroutine.
```
disruptek, line 332: Needs a `,` and `s/pass in/pass it in/` but this material seems to be out of place; the `f` application was two fences ago and interrupted by the `cpsVoodoo` explanation.  I think maybe the voodoo treatment has to move up.

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+
+```nim
+dest.send(n)
+```
+
+Again, the *sending* is just updating the state of the continuation and resuming
+it from `dest.cResume`. *Receiving*, then done by `consumer`, is just getting that
+data from the continuation.
+
+Setting the `value` to the received integer we're finally able to print it:
```
disruptek, line 342: _After setting_ and _integer, we're_

README.md
```diff
@@ -219,9 +219,137 @@ echo "it's later!  time to count!"
...
+
+```nim
+let value = recv()
+echo value
+```
+
+Since we're at the end of the code of both coroutines, their loops bring us to 
+yielding again, now in the opposite order: first the `consumer` and right next 
+the `filter` suspend execution which returns us to the main loop, ready to go
+again.
```
disruptek, line 352: `s/right//` and `s/bring us to yielding/yield/` or somesuch simplification.  _next, the `filter` suspends, which returns_ might be easier to read, also.

disruptek (commented):
Really happy to see this addition to the docs; I've been lazy about it.  Thank you!

README.md
```diff
@@ -254,12 +254,12 @@ Both coroutines are already "running" (`coroX.running == true`) but nothing is
...
 ```nim
-coro1.resume()
-coro2.resume()
+discard coro1.resume()
+discard coro2.resume()
 ```
 
 The `resume` proc is our dispatcher for the coroutines and it's what actually
-calls the execution via the function pointer inside the continuation:
+runs the continuation via the call to a function pointer held by continuation:
```
disruptek, line 262: _via the call to the function pointer stored in the continuation_ or similar?

README.md
```diff
@@ -269,8 +269,13 @@ proc resume(c: Coroutine): Coroutine {.discardable.} =
   result = Coroutine c
 ```
 
-Before invoking that function we need to convert the Coroutine to its base class
-first. Then, the filter coroutine gets launched:
+Actually, this is such a basic pattern for CPS code that the library provides a
+[trampoline](https://nim-works.github.io/cps/cps/spec.html#trampoline%2CsinkT)
+template for this, and it would be preferable to just use it here. Also, notice
+how before invoking the function pointer we needed to convert a `Coroutine` to
+its base type so we could set `c` to `fn(c)`, which returns a `Continuation`.
```
disruptek, line 276: This is excellent.

README.md
```diff
@@ -281,24 +286,27 @@ proc filter(dest: Coroutine, f: proc(x: int): int) {.cps:Coroutine.} =
 ```
 
 As the first coroutine launches, it yields (suspends execution) by calling
-`jield` immediately (*yeild* is a keyword in Nim). This is necessary because we
+`jield` immediately (*yield* is a keyword in Nim). This is necessary because we
 haven't actually sent any data to process, yet we need the Coroutines launched
 and waiting for it. We could move the suspension points later, but we don't want
 the coroutines processing the initialized-by-default value prior to the data we
-send them.
+send them (try moving "jields" around and inspect the results).
```
disruptek, line 293: I love it.  We really need more of these calls to action.

examples/coroutine.nim
```diff
@@ -18,12 +18,12 @@ proc recv(c: Coroutine): int {.cpsVoodoo.} =
 
 # Suspend execution of the coroutine
 proc jield(c: Coroutine): Coroutine {.cpsMagic.} =
-  c.cResume = c
+  c.next = c
```
disruptek, line 21: Perfect.

disruptek (changes requested):
Damn near perfect.  Nice job.

ZoomRmc:
Thanks a lot for your review, @disruptek, it's very helpful as you accurately point to places where my understanding still lacks.

The explanation of `resume` and `jield` could still to be improved further but it's better than nothing.

PR #262, opened 2021-12-29
rename `()` to `recover`

disruptek:
Also improves the trace output a tiny bit.


alaviss (approved):
Should probably elaborate that the change is done due to `()` causing weird Nim issues, lgtm otherwise.

PR #280, opened 2022-12-01
restoring the call operator in nim-1.7.3+

disruptek:
This will need a rebase; I'm lazy.


metagn:
1.6.11 should also be fine (https://github.com/nim-lang/Nim/commit/0683e8f747b3f5605333cbb25884bc90ebaa07dc). Also thank you very much.

PR #283, opened 2022-12-05
Mark the Continuation type as .acyclic. to keep orcs hands off it

zevv:
this should allow CPS to be used with orc.

This does need some more discussion before merging...

### DIFF
--- a/cps/spec.nim
+++ b/cps/spec.nim
@@ -57,7 +57,7 @@ const
   traceDequeSize* {.intdefine, used.} = 4_096
 
 type
-  Continuation* = ref object of RootObj
+  Continuation* {.acyclic.} = ref object of RootObj
     fn*: proc(c: Continuation): Continuation {.nimcall.} ##
     ## The `fn` points to the next continuation leg.
     mom*: Continuation  ##


zevv:
Of course better would be if we can get rid if the cycle altogether, but I still don't properly understand the details of this.

disruptek:
Can you prove we never need a cycle?  What about making an acyclic subtype?

This doesn't feel like an easy guarantee to break, nor does it feel like an easy guarantee to make.  😒

alaviss:
> Of course better would be if we can get rid if the cycle altogether, but I still don't properly understand the details of this.

AFAIK the only "intentional" cycle we make in the library is `child -> mom`, `mom -> child` where the former is required to return control and the latter is used to obtain procedure results.

I don't see a good way to get rid of the former but the latter can potentially be removed if we have some form of "message passing" (ie. a way to set variables in a continuation outside of it).

We can potentially side-step this issue altogether if we store the continuation "stack" in an array or something, but then that'd require either analysis to predetermine the maximum "stack" size, or a seq which have its own set of problems.

Then, of course, we still have the potential of intentional cycle by the user, which is a pain on it's own, although we can just say "don't make cycles", I guess?

disruptek:
We should move to result pointers to solve problems with result recovery.  This coincidentally solves _some_ cycles, but really, I think we should stop caring so much about cycles.  They are only a problem for technical reasons and not semantic ones.  Better to design for cycles, instead.

alaviss:
I think a way to pass data in-out on resume/suspend points would be pretty neat. I can't think of /how/ to achieve that safely, though.

The idea is that if we have "native" data passing, we don't need the current tricks for result, and suspending magic can actually "return" value without pointer/continuation shenanigans.

disruptek:
What does "safely" mean in this context?

Suspend/resume points just means continuation legs, to me, so I see a number of options once we accept that the trampoline may have to change yet again.

It probably needs to be implemented as a core macro.

alaviss:
> What does "safely" mean in this context?

Some kind of type-safety, either by clever use of generics or simply RTTI.

disruptek:
Clearly, this isn't the solution.

PR #285, opened 2022-12-09
small improvements to memory behavior

disruptek:
- changed where we nil the mom
- added destructor hooks for testing
- changed the trampoline to move more


cps/spec.nim
```diff
@@ -97,6 +98,13 @@ type
     Tail    = "tail"      ## invoked when a new continuation has a parent
     Stack   = "stack"     ## invoked to annotate stack semantics
 
+proc `=copy`(dest: var ContinuationObj; src: ContinuationObj) {.error.} =
+  discard
```
alaviss, line 102: IIRC you don't need an impl when its `{.error.}`

cps/spec.nim
```diff
@@ -97,6 +98,13 @@ type
...
+
+proc `=destroy`(dest: var ContinuationObj) =
+  for key, value in dest.fieldPairs:
+    reset value
```
alaviss, line 106: I'm pretty sure this is the default behavior, so maybe we don't need it.

If we do it's probably an upstream bug.

cps/spec.nim
```diff
@@ -485,10 +493,12 @@ proc writeTraceDeque*(c: Continuation) {.cpsVoodoo.} =
...
   ## This is the basic trampoline: it will run the continuation
   ## until the continuation is no longer in the `Running` state.
-  var c: Continuation = c
+  var c: Continuation = move c
   while not c.isNil and not c.fn.isNil:
     try:
-      c = c.fn(c)
+      var x = c.fn(c)
+      wasMoved c
+      c = move x
```
alaviss, line 501: Do we really have to do all this? This looks like default behavior to me.

We can probably throw Nim a bone with `var x`, but we don't need `wasMoved` (which is just an alias for `reset` IIRC) for sure.

cps/spec.nim
alaviss: I'd also say that `=destroy` doesn't execute reliably on non-ARC, so I'm a tiny bit concerned about using it while we still leave non-ARC builds open.

alaviss (commented):
Just some quick thoughts

cps/spec.nim
disruptek: It's just semantic sugar, but I don't want the destroy.

cps/spec.nim
disruptek: We don't support non-arc.  To put that another way, we should not.  And no one should be compiling with refc, let alone cps with refc.

disruptek: Also, the main reason to put this in (now) is so we can use it in debugging and eventual trace events.

cps/spec.nim
alaviss: Destroy behavior is weird. It almost always called since the "spec" is very lax on it and no amount of code can get rid of it (unless you go `{.nodestroy.}` on the proc, but that is another can of worms). I had to give up on getting them not called in nim-sys.

alaviss: I guess what I'm saying is, if you don't want them to trigger your hooks, you need to figure out a "canonical zero" state to ignore destructors calls. Otherwise it will almost always be called if there's a non-zero chance of it being necessary.

disruptek: I will see about adding some tests with the killer.  I want it to be called and when debugging, it seems to be called appropriately.  But the tests will help ensure this stays true.

PR #292, opened 2023-04-24
recursive dependency: 'cpsJump'

Gruruya:
Caused by https://github.com/nim-lang/Nim/commit/63d29ddd6980ee9f89673c454c15da52e2984283 PR: https://github.com/nim-lang/Nim/pull/21671
Without this patch, attempting to build [disruptek/balls@#v4](https://github.com/disruptek/balls/tree/v4) gives
```shell
cps/transform.nim(70, 19) Error: recursive dependency: 'cpsJump'
Error: Build failed for the package: balls
```
https://github.com/nim-works/cps/blob/5645eb5838e6d30d484fbe23d8c590201e520630/cps/transform.nim#L70

### DIFF
--- a/cps/spec.nim
+++ b/cps/spec.nim
@@ -244,7 +244,7 @@ template rewriteIt*(n: typed; body: untyped): NormNode =
   body
   workaroundRewrites it
 
-template debugAnnotation*(s: typed; n: NimNode; body: untyped) {.dirty.} =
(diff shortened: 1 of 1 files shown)


disruptek:
I'm not in favor of giving up `typed` code for `untyped` code just to chase _devel_.  I will submit an issue and let's see if we can simply have the bug fixed in the compiler instead.  It's not like there's a bunch of code that relies upon this defect, right...

PR #294, opened 2023-05-14
Remove CallNodes/nnkCallKinds comparison; fixes #293

### LINKED ISSUE #293: CallNodes = nnkCallKinds
https://github.com/nim-works/cps/blob/5645eb5838e6d30d484fbe23d8c590201e520630/cps/transform.nim#L9-L10

https://github.com/nim-lang/Nim/pull/21833
https://github.com/nim-lang/Nim/pull/21829

Gruruya:
Could change the comparison, but it seems to me like we can expect for them to remain equal

### DIFF
--- a/cps/transform.nim
+++ b/cps/transform.nim
@@ -6,9 +6,6 @@ export Continuation, ContinuationProc, cpsCall, cpsMustJump
 
 #{.experimental: "strictNotNil".}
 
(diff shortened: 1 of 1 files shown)


Gruruya:
Can't reproduce the CI failure on my machine so I don't know whats causing it.  
It's not this PR anyway and making this change at least gets balls compiling.

PR #296, opened 2023-05-24
Adjust for `nnkCallKinds` == `CallNodes`

### LINKED ISSUE #293: CallNodes = nnkCallKinds

Gruruya:
`nnkHiddenCallConv` was added to `nnkCallKinds`, so we need to shift some code. fixes #293

### DIFF
--- a/cps.nim
+++ b/cps.nim
@@ -111,7 +111,7 @@ proc doWhelp(n: NormNode; args: seq[NormNode]): Call =
 template whelpIt*(input: typed; body: untyped): untyped =
   ## Instantiate the given continuation call and inject `it` in the body.
   var n = normalizeCall input
-  if n.kind in (CallNodes - {nnkHiddenCallConv}):
+  if n.kind in NormalCallNodes:
     var it {.inject.} = doWhelp(n[0], n[1..^1])
     body
     NimNode it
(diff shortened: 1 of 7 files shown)


Gruruya:
May have been a red herring, or at least it's not the thing failing tests. Seems that https://github.com/nim-lang/Nim/commit/c2bcfd8cd908265c358c60c4da137783b10a8549 is the real offender.

disruptek:
Should we just add a new symbol for `CallNodes - {nnkHiddenCallConv}` so we don't have to make these changes everywhere next time?

Gruruya:
Nice job

disruptek:
> I kept nnkCallNodes in normalizedast, but I'm not sure if that's right or if there's any other instances that should be kept as nnkCallNodes as well

Well, where?

Gruruya:
`abbreviation` particularly

disruptek:
No big deal; the tests pass and we can add more when it fails.  :shrug:

disruptek:
Nice to have this working again, thanks!

PR #299, opened 2023-05-25
towards unbreaking nim-devel yet again

disruptek:


disruptek:
The problem at this point is the `_` symbols which receive the RHS values as in
```nim
(a, b) = (1, 2)
```
which is rewritten to
```nim
env.a = __2394223423[0]
env.b = __2394223423[1]
```
and after tracking down most of the symbols and giving them line info, these remain unwritten.

PR #322, opened 2024-02-25
perform a missing dealloc() in bootstrap

disruptek:
Unfortunately, this is really just a partial solution to making alloc/dealloc work more widely.  The hook might just be poorly conceived.

### DIFF
--- a/cps/environment.nim
+++ b/cps/environment.nim
@@ -576,24 +576,22 @@ proc createBootstrap*(env: Env; n: ProcDef, goto: NormNode): ProcDef =
   elif env.rs.hasType:
     result.body.add:
       # then at runtime, issue an if statement to
-      nnkIfExpr.newTree:
+      nnkIfStmt.newTree:
         nnkElifExpr.newTree(
           # check if the continuation is not nil, and if so, to
           newCall(bindSym"not", newDotExpr(c, asName"dismissed")),
           # assign the result from the continuation's result field
           newAssignment(asName"result",
                         newCall(ident"move",
-                        newDotExpr(env.castToChild(c), env.rs.name)).NormNode)
-        )
+                        newDotExpr(env.castToChild(c), env.rs.name)).NormNode))
   # perform a dealloc() on the continuation if we can
   result.body.add:
-    nnkIfExpr.newTree:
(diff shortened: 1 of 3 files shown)


cps/environment.nim
```diff
@@ -582,8 +582,18 @@ proc createBootstrap*(env: Env; n: ProcDef, goto: NormNode): ProcDef =
           newCall(bindSym"not", newDotExpr(c, asName"dismissed")),
           # assign the result from the continuation's result field
           newAssignment(asName"result",
-            newDotExpr(env.castToChild(c), env.rs.name))
+                        newCall(ident"move",
+                        newDotExpr(env.castToChild(c), env.rs.name)).NormNode)
```
alaviss, line 586: Maybe use `newAssignment(asName"result"): ...` instead so that it looks a little clearer.

cps/environment.nim
```diff
@@ -582,8 +582,18 @@ proc createBootstrap*(env: Env; n: ProcDef, goto: NormNode): ProcDef =
...
         )
+  # perform a dealloc() on the continuation if we can
+  result.body.add:
+    nnkIfExpr.newTree:
+      nnkElifExpr.newTree(
+        # check if the continuation is not nil, and if so, to
+        newCall(bindSym"not", newDotExpr(c, asName"dismissed")),
+        # apply the dealloc() hook on the continuation
+        newAssignment(c, Dealloc.hook(env.castToRoot(c), env.identity))
+      )
```
alaviss, line 596: It's probably better to use `nnkIfStmt`. `Expr` might break things since parser only produce this for `if x: y else: z` iirc.

```suggestion
    nnkIfStmt.newTree:
      nnkElifBranch.newTree(
        # check if the continuation is not nil, and if so, to
        newCall(bindSym"not", newDotExpr(c, asName"dismissed")),
        # apply the dealloc() hook on the continuation
        newAssignment(c, Dealloc.hook(env.castToRoot(c), env.identity))
      )
```

alaviss (approved):
Seems alright to me, but I have a few suggestions below

disruptek:
Yeah, I did try to improve the syntax but the compiler won this time.

PR #331, opened 2024-08-23
cps: implement shims for docgen

alaviss:
Currently CPS simply left the procedures untouched during docgen. This causes problems when code for `whelp` gets sem-ed, breaking docgen.

This commit implements a couple docgen shims:

- `cps` now result in a `{.cps: baseType.}` pragma tagged to the resulting procedure and will show up in the generated documentation.
- `whelp` will generate `nil` continuation for the base type of a cps proc.
- `whelp` will generate a `Callback` with a factory type that yields a the base continuation type, not the environment type. This should work for most cases when docgen are concerned.

This worked well enough for `insideout` docs to be generated successfully under NimSkull.


alaviss:
How the `cps` tag looks like in docgen. From `sys/sockets`.

image

